PR #8, opened 2023-01-24
Fix issues discovered in initial deployment.

sharkinsspatial:
Several small logic errors and an unsolved issue involving Lambda role assumption were discovered during the initial attempt at deploying the `cdk-pgstac` construct via https://github.com/developmentseed/aws-asdi-pgstac. The included fix for using a broad role for public buckets is temporary and we'll need to diagnose the issue we are seeing with `assume_role` going forward.


alukach:
If it's okay with you, I'm going to avoid reviewing this and we can instead discuss & merge in these changes in #10

PR #18, opened 2023-03-06
feat: collection endpoint

jjfrench:
- Porting over changes from https://github.com/NASA-IMPACT/veda-stac-ingestor/pull/29
- Adding basic tests to the collection endpoint

### DIFF
--- a/lib/ingestor-api/index.ts
+++ b/lib/ingestor-api/index.ts
@@ -36,6 +36,10 @@ export class StacIngestor extends Construct {
       env,
       dataAccessRole: props.dataAccessRole,
       stage: props.stage,
+      dbSecret: props.stacDbSecret,
+      dbVpc: props.vpc,
+      dbSecurityGroup: props.stacDbSecurityGroup,
+      subnetSelection: props.subnetSelection,
     });
 
     this.buildApiEndpoint({
--- a/lib/ingestor-api/index.ts
+++ b/lib/ingestor-api/index.ts
@@ -84,6 +84,10 @@ export class StacIngestor extends Construct {
     env: Record<string, string>;
     dataAccessRole: iam.IRole;
     stage: string;
+    dbSecret: secretsmanager.ISecret;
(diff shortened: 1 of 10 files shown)


sharkinsspatial (changes requested):
@jjfrench Your branch fails when deploying to the testing stack for `aws-asdi-pgstac` with the following error 
```
Resource handler returned message: "The provided execution role does not have permissions to call Create
NetworkInterface on EC2 (Service: Lambda, Status Code: 400, Request ID: 077eece3-808c-4d43-9bec-b6149d83
6196)" (RequestToken: d8e06c8a-dac2-a260-f97b-4a29700a5903, HandlerErrorCode: InvalidRequest)
```
@alukach can you review this PR as well and @jjfrench can you coordinate with the `veda-ingestor` folks who developed this to confirm what additional changes we need to make to the ingestor handler role?  I'll also try have a look at the codebase there tomorrow and see what might be the issue.

jjfrench:
@sharkinsspatial 
> I was able to fix the deployment issue by adding a managed policy to the Lambda execution role via [f15e982](https://github.com/developmentseed/cdk-pgstac/commit/f15e982adc685c95ac20fb835218277b3f9de484).
> 
> But now it appears there is a VPC connectivity issue between the ingestor api handler and the pgSTAC RDS instance as a POST to the `/collections` endpoint results in a Lambda timeout.

The VPC connectivity issue is also blocking deployment or where is this deployed at so I can take a look? We just had similar issues to the one described above when deploying the veda resources on MAAP, but I didn't encounter this issue when deploying a test stack to my personal aws account with the same vpc configurations 🤔

sharkinsspatial:
@jjfrench As discussed in the `earthdata-team` Slack thread, deployment is ok, but the api handler Lambda will timeout when trying to connect to the RDS instance in a private subnet.  I suspect that the deployment issue only related to altering a network interface and so only occurs when updating an existing stack which is why you might be seeing this on MAAP but not a fresh deployment.

PR #24, opened 2023-03-16
chore: rename file & class to make project-agnostic

alukach:
Let's make this code project-agnostic

Is this a breaking change? What should the impact be on the project versioning?


vincentsarago (approved):

vincentsarago:
> Is this a breaking change? What should the impact be on the project versioning?

It's not an API changed and will not impact how people use the library so I guess a minor update should be more than enough

PR #26, opened 2023-03-22
fix(ingestor-api)!: store STAC item as string in DynamoDB

sharkinsspatial:
This PR addresses the issue raised in https://github.com/NASA-IMPACT/veda-data-pipelines/issues/313.  The PR updates the ingestor code to serialize the STAC `item` property of the `ingestion` as a string prior to storing it in DynamoDB.  Additionally, it updates the code used to deserialize `ingestions` queried from the DynamoDB table and load them into `pgstac` and adds unit test coverage for some of the `ingestor` code paths (which previously had no coverage).

### DIFF
--- a/lib/ingestor-api/runtime/src/schemas.py
+++ b/lib/ingestor-api/runtime/src/schemas.py
@@ -3,10 +3,10 @@ import binascii
 import enum
 import json
 from datetime import datetime
-from decimal import Decimal
 from typing import TYPE_CHECKING, Dict, List, Optional, Union
 from urllib.parse import urlparse
 
+from fastapi.encoders import jsonable_encoder
 from fastapi.exceptions import RequestValidationError
 from pydantic import (
     BaseModel,
@@ -30,7 +30,7 @@ class AccessibleAsset(shared.Asset):
         url = urlparse(href)
 
         if url.scheme in ["https", "http"]:
-            validators.url_is_accessible(href)
+            validators.url_is_accessible(href=href)
         elif url.scheme in ["s3"]:
             validators.s3_object_is_accessible(
                 bucket=url.hostname, key=url.path.lstrip("/")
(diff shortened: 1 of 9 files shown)


lib/ingestor-api/runtime/src/schemas.py
```diff
@@ -73,19 +74,6 @@ class Ingestion(BaseModel):
...
-            # NOTE: when serializing, must call as ingestion.json(models_as_dict=False)
-            Item: lambda item: item.json(by_alias=True),
-        }
-
-    def json(self, *args, **kwargs):
-        # Update default to not represent models (e.g. `items` property) as a dict to
-        # allow our `json_encoders` override to properly serialize `items` property
-        kwargs.setdefault("models_as_dict", False)
-        return super().json(*args, **kwargs)
-
```
alukach: @sharkinsspatial My rationale here is that if we need to put all this cruft in to make special considerations for serializing our `Ingestion` object for DynamoDB, it would probably make for cleaner code if we just put that cruft into the `dynamodb_dict()` method.

lib/ingestor-api/runtime/src/schemas.py
```diff
@@ -73,19 +73,6 @@ class Ingestion(BaseModel):
...
```
alukach: @sharkinsspatial My rationale here is that if we need to put all this cruft in to make special considerations for serializing our `Ingestion` object for DynamoDB, it would probably make for cleaner code if we just put that cruft into the `dynamodb_dict()` method.

alukach (approved):
Looks great @sharkinsspatial!  Thanks for adding those tests.  I added a few more (once some are written, it makes it so much easier to expand).

I rewrote a bit of the code to make use of `fastapi.encoders.jsonable_encoder()` which I think makes for slightly cleaner code.

Feeling really good about this change!

PR #34, opened 2023-04-06
fix(bootstraper): fix version inconsistencies and #19 regressions

emileten:
- [x] fix inconsistent `pypgstac` versions in bootstrapper and stac-ingestor
- [x] fix regression introduced by #19 : `pypgstac.load` does not have a `load_collection` but a `load_collections` method. I am unsure about my fix : it looks like `load_collections` excepts a json containing a list of collection metadata. How is it going to behave if we have a single collection in it ? 
- [x] make sure the value passed to the `file` parameter of the `pypgstac` collection loader is a `list` of `dict`.

### DIFF
--- a/lib/bootstrapper/index.ts
+++ b/lib/bootstrapper/index.ts
@@ -17,7 +17,7 @@ function hasVpc(
   return (instance as aws_rds.DatabaseInstance).vpc !== undefined;
 }
 
-const DEFAULT_PGSTAC_VERSION = "0.6.8";
+const DEFAULT_PGSTAC_VERSION = "0.6.13";
 
 /**
  * Bootstraps a database instance, installing pgSTAC onto the database.
--- a/lib/ingestor-api/runtime/src/collection.py
+++ b/lib/ingestor-api/runtime/src/collection.py
@@ -18,6 +18,7 @@ def ingest(collection: StacCollection):
         creds = get_db_credentials(os.environ["DB_SECRET_ARN"])
         with PgstacDB(dsn=creds.dsn_string, debug=True) as db:
             loader = Loader(db=db)
+            collection = [collection.to_dict()] # pypgstac wants either a string or an Iterable of dicts. 
             loader.load_collections(file=collection, insert_mode=Methods.upsert)
     except Exception as e:
         print(f"Encountered failure loading collection into pgSTAC: {e}")
--- a/lib/ingestor-api/runtime/src/collection.py
+++ b/lib/ingestor-api/runtime/src/collection.py
@@ -18,7 +18,9 @@ def ingest(collection: StacCollection):
         creds = get_db_credentials(os.environ["DB_SECRET_ARN"])
         with PgstacDB(dsn=creds.dsn_string, debug=True) as db:
             loader = Loader(db=db)
-            collection = [collection.to_dict()] # pypgstac wants either a string or an Iterable of dicts. 
+            collection = [
+                collection.to_dict()
+            ]  # pypgstac wants either a string or an Iterable of dicts.
             loader.load_collections(file=collection, insert_mode=Methods.upsert)
     except Exception as e:
         print(f"Encountered failure loading collection into pgSTAC: {e}")
--- a/lib/ingestor-api/runtime/src/collection.py
+++ b/lib/ingestor-api/runtime/src/collection.py
@@ -18,7 +18,7 @@ def ingest(collection: StacCollection):
         creds = get_db_credentials(os.environ["DB_SECRET_ARN"])
         with PgstacDB(dsn=creds.dsn_string, debug=True) as db:
             loader = Loader(db=db)
-            loader.load_collection(file=collection, insert_mode=Methods.upsert)
+            loader.load_collections(file=collection, insert_mode=Methods.upsert)
     except Exception as e:
         print(f"Encountered failure loading collection into pgSTAC: {e}")
 
--- a/lib/ingestor-api/runtime/tests/conftest.py
+++ b/lib/ingestor-api/runtime/tests/conftest.py
@@ -248,7 +248,7 @@ def client_authenticated(app):
     """
     from src.dependencies import get_username
 
-    app.dependency_overrides[get_username] = lambda: 'test_user'
+    app.dependency_overrides[get_username] = lambda: "test_user"
     return TestClient(app)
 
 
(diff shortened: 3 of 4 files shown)


vincentsarago (approved):

lib/bootstrapper/index.ts
```diff
@@ -17,7 +17,7 @@ function hasVpc(
   return (instance as aws_rds.DatabaseInstance).vpc !== undefined;
 }
 
-const DEFAULT_PGSTAC_VERSION = "0.6.8";
+const DEFAULT_PGSTAC_VERSION = "0.6.13";
```
alukach, line 20: https://github.com/developmentseed/cdk-pgstac/blob/d6ce6005d5e7f72d2c32dc98659096385c079df7/lib/ingestor-api/runtime/requirements.txt#L9

lib/ingestor-api/runtime/src/collection.py
```diff
@@ -18,7 +18,7 @@ def ingest(collection: StacCollection):
         creds = get_db_credentials(os.environ["DB_SECRET_ARN"])
         with PgstacDB(dsn=creds.dsn_string, debug=True) as db:
             loader = Loader(db=db)
-            loader.load_collection(file=collection, insert_mode=Methods.upsert)
+            loader.load_collections(file=collection, insert_mode=Methods.upsert)
```
alukach, line 21: ✅ 
https://github.com/stac-utils/pgstac/blob/v0.6.13/pypgstac/pypgstac/load.py#L188

alukach (approved):

emileten:
I tested this in an AWS deployment. I am posting this to the collections endpoint : 

```
{
    "id": "icesat2-boreal",
    "links": [
        {
          "rel": "items",
          "type": "application/geo+json",
          "href": "https://stac.maap-project.org/collections/icesat2-boreal/items"
        },
        {
          "rel": "parent",
          "type": "application/json",
          "href": "https://stac.maap-project.org/"
        },
        {
          "rel": "root",
          "type": "application/json",
          "href": "https://stac.maap-project.org/"
        },
        {
          "rel": "self",
          "type": "application/json",
          "href": "https://stac.maap-project.org/collections/icesat2-boreal"
        }
    ],
    "type": "Collection",
    "title": "Gridded Boreal Aboveground Biomass Density c.2020 at 30m resolution",
    "extent": {
        "spatial": {
            "bbox": [
                [
                    -180, 51.6, 180, 78
                ]
            ]
        },
        "temporal": {
            "interval": [
              [
                "2019-01-01T00:00:00.000Z",
                "2021-01-01T00:00:00.000Z"
              ]
            ]
          }        
    },
    "license": "CC-BY",
    "description": "Aboveground biomass density c.2020 gridded to 30m derived from ICESat-2, Harmonized Landsat-Sentinel2 and Copernicus DEM. Band 1 is the data band. Band 2 is the standard deviation.",
    "item_assets": {
        "cog_default": {
            "type": "image/tiff; application=geotiff; profile=cloud-optimized",
            "roles": [
                "data",
                "layer"
            ],
            "title": "Default COG Layer",
            "description": "Cloud optimized default layer to display on map"
        },
        "documentation": {
            "type": "text/html",
            "roles": [
                "documentation"
            ],
            "title": "Find and open data with STAC",
            "description": "Find and open data with STAC",
            "href": "https://docs.maap-project.org/en/stac/search/searching_the_stac_catalog.html"
        }        
    },
    "stac_version": "1.0.0"
}
```

And I am getting an error in the database side. Here is the log :

```
2023-04-07 07:10:37 UTC:10.0.2.183(22885):pgstac_user@pgstac:[4632]:ERROR: COPY from stdin failed: error from Python: JSONDecodeError - Input must be bytes, bytearray, memoryview, or str: line 1 column 1 (char 0)
```

@vincentsarago @alukach any idea ? This PR is bumping the database `pypgstac` version so that it matches that of the ingestor -- so it bumps from 0.6.8 to 0.6.13. Could this update be causing the error ? 

Also @alukach -- do you have a way to test changes in a deployment _before_ publishing them ? I have been trying to point [my package.json file](https://github.com/MAAP-Project/maap-cdk-pgstac) to a github branch or local cdk-pgstac directory, without success so far. I keep encountering Typescript compiling errors. I am wondering how you typically do this. Thanks !

sharkinsspatial:
@emileten Given some oddities with `npm` I have found this is the only reliable approach for testing the compiled construct in another stack.  

- Update the package version (this is a temporary hack to avoid potential issues with npm caching, make sure not to commit this change) https://github.com/developmentseed/cdk-pgstac/blob/main/package.json#L3
- `npm run build` and `npm run package`
- Update the `pacakge.json` in the target CDK project where you are using the `cdk-pgstac` construct to reference your new version package `"cdk-pgstac": "file:../cdk-pgstac/dist/js/cdk-pgstac@your_version.jsii.tgz"`.
- `rm -rf node_modules` in the target CDK project.
- `npm install` in the target CDK project.

I'm not certain all this is fully necessary, but I ran into numerous cases where my target CDK project was using a cached version of the locally compiled package so I've been using this approach to ensure I'm always using the compiled code I expect.

sharkinsspatial:
cc @bitner An ideas on why `pgstac` would be throwing this https://github.com/developmentseed/cdk-pgstac/pull/34#issuecomment-1500018885?

bitner:
> cc @bitner An ideas on why `pgstac` would be throwing this [#34 (comment)](https://github.com/developmentseed/cdk-pgstac/pull/34#issuecomment-1500018885)?

You say "posting this to the endpoint". Is this being posted to a stac-fastapi endpoint or to something that is using pypgstac? For pypgstac, the read_json function (https://github.com/stac-utils/pgstac/blob/64221059916d4e182c938c86799bf77d9c403965/pypgstac/pypgstac/load.py#L145) expects either a json string or an iterable of python dicts `[{"id": .....}]`

sharkinsspatial:
@emileten This looks good to me and nice catch.  This is a great example of why we need https://github.com/developmentseed/cdk-pgstac/issues/13 (which would have caught this regression which actually got introduced across 2 PRs) as well as configuring type checking in our `tox` configuration.

Can you add a unit test for https://github.com/developmentseed/cdk-pgstac/blob/main/lib/ingestor-api/runtime/src/collection.py which validates that `load_collections` is being called with the wrapped iterable.  You can use https://github.com/developmentseed/cdk-pgstac/blob/main/lib/ingestor-api/runtime/tests/test_ingestor.py as an example of the mocking.

PR #35, opened 2023-04-06
feat(ingestor): validation exception handler

emileten:
- [x] add handler for ingestor API invalid request exceptions, so that more informative errors are returned to users of the API
- [x] modify ingestor tests accordingly
- [x] make `item_assets` optional in the Collection schemas.

### DIFF
--- a/lib/ingestor-api/runtime/src/main.py
+++ b/lib/ingestor-api/runtime/src/main.py
@@ -1,4 +1,6 @@
 from fastapi import Depends, FastAPI, HTTPException
+from fastapi.exceptions import RequestValidationError
+from fastapi.responses import JSONResponse
 
 from . import collection as collection_loader
 from . import config, dependencies, schemas, services
@@ -123,3 +125,8 @@ def who_am_i(username=Depends(dependencies.get_username)):
     Return username for the provided request
     """
     return {"username": username}
+
+# exception handling
+@app.exception_handler(RequestValidationError)
+async def validation_exception_handler(request, exc):
+    return JSONResponse(str(exc), status_code=422)
--- a/lib/ingestor-api/runtime/src/main.py
+++ b/lib/ingestor-api/runtime/src/main.py
@@ -126,6 +126,7 @@ def who_am_i(username=Depends(dependencies.get_username)):
     """
     return {"username": username}
 
+
 # exception handling
 @app.exception_handler(RequestValidationError)
 async def validation_exception_handler(request, exc):
--- a/lib/ingestor-api/runtime/src/schemas.py
+++ b/lib/ingestor-api/runtime/src/schemas.py
@@ -52,7 +52,7 @@ class AccessibleItem(Item):
 
 class StacCollection(Collection):
     id: str
-    item_assets: Dict
+    item_assets: Optional[Dict]
 
 
 class Status(str, enum.Enum):
--- a/lib/ingestor-api/runtime/tests/conftest.py
+++ b/lib/ingestor-api/runtime/tests/conftest.py
@@ -248,7 +248,7 @@ def client_authenticated(app):
     """
     from src.dependencies import get_username
 
-    app.dependency_overrides[get_username] = lambda: 'test_user'
+    app.dependency_overrides[get_username] = lambda: "test_user"
     return TestClient(app)
 
 
--- a/lib/ingestor-api/runtime/tests/test_registration.py
+++ b/lib/ingestor-api/runtime/tests/test_registration.py
@@ -119,7 +119,9 @@ class TestCreate:
         )
         assert response.status_code == 422, "should get validation error"
         for asset_type in self.example_ingestion.item.assets.keys():
-            assert asset_type in response.json(), "should reference asset type in validation error response"
+            assert (
+                asset_type in response.json()
+            ), "should reference asset type in validation error response"
         assert (
             len(self.db.fetch_many(status="queued")["items"]) == 0
         ), "data should not be stored in DB"
--- a/lib/ingestor-api/runtime/tests/test_registration.py
+++ b/lib/ingestor-api/runtime/tests/test_registration.py
@@ -119,12 +119,7 @@ class TestCreate:
         )
         assert response.status_code == 422, "should get validation error"
         for asset_type in self.example_ingestion.item.assets.keys():
-            assert any(
-                [
-                    err["loc"] == ["body", "assets", asset_type, "href"]
-                    for err in response.json()["detail"]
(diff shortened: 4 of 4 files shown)


vincentsarago (approved):

lib/ingestor-api/runtime/src/schemas.py
```diff
@@ -52,7 +52,7 @@ def exists(cls, collection):
 
 class StacCollection(Collection):
     id: str
-    item_assets: Dict
+    item_assets: Optional[Dict]
```
alukach, line 55: Something feels a bit off about this.  Either we're using the [Item Assets extension](https://github.com/stac-extensions/item-assets#collection-fields) or we're not.

Perhaps we should put this behind a feature flag?  Not really critical for this PR though.

lib/ingestor-api/runtime/src/main.py
```diff
@@ -123,3 +125,9 @@ def who_am_i(username=Depends(dependencies.get_username)):
     Return username for the provided request
     """
     return {"username": username}
+
+
+# exception handling
+@app.exception_handler(RequestValidationError)
+async def validation_exception_handler(request, exc):
+    return JSONResponse(str(exc), status_code=422)
```
alukach, line 133: What's the backstory for this change?

Can you provide an example of the new error format compared to the old format?  My hunch is that this makes it so that we are returning a raw string to the user?

alukach (changes requested):
I am not familiar with why this change is desired.  Can you provide more of the backstory?

lib/ingestor-api/runtime/src/schemas.py
emileten: I thought we could make it optional since it's not required in the base STAC specification (didn't realize it's an extension in itself). In MAAP we have a couple collection files without this information. I agree feature flags sounds like a better way to handle this.

lib/ingestor-api/runtime/src/main.py
emileten: sorry I should have provided more context. It's an improvement they made in the VEDA version of the ingestor. My understanding is that it makes sure to return validation information when there is a validation error. 

For example, the collection endpoint. If you pass an invalid collection json (invalid with respect to the model described in schemas), then you will see an error indicating what's missing etc.

As far as I experienced, it's not the case without this handler. You are right that with this we're returning a raw string to the user.

emileten: @alukach here the change in practice. Say I have a collection JSON that doesn't have a `links` field, and I submit that to the collection endpoint. 

I thought their change was actually providing more information but it's not the case, it just changes the format of the error. I think it's more readable, so that would be the improvement of this change. Let me know if you think it worths the addition. 

Client side error before the change : 

```
Error inserting collection.
{"detail":[{"loc":["body","links"],"msg":"field required","type":"value_error.missing"}]}
  File "/Users/emiletenezakis/devseed/maap-pgstac/test_insert_collection.py", line 145, in <module>
    response.raise_for_status()
  File "/Users/emiletenezakis/mambaforge/lib/python3.10/site-packages/requests/models.py", line 1021, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 422 Client Error: Unprocessable Entity for url: http://127.0.0.1:8000/collections?provided_by=emile+tenezakis
```

Client side error after the change : 

```
Error inserting collection.
"1 validation error for Request\nbody -> links\n  field required (type=value_error.missing)"
  File "/Users/emiletenezakis/devseed/maap-pgstac/test_insert_collection.py", line 145, in <module>
    response.raise_for_status()
  File "/Users/emiletenezakis/mambaforge/lib/python3.10/site-packages/requests/models.py", line 1021, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 422 Client Error: Unprocessable Entity for url: http://127.0.0.1:8000/collections?provided_by=emile+tenezakis
```

alukach: To be honest, I'm not sure if this is something we should port over.  This change makes the API response more _human_ readable, but more difficult to parse via machine.  In my experience, APIs typically don't serve humans and instead serve client applications.  Building logic into a frontend client that highlights the field in the form that caused the validation error would be considerably more difficult to categorize and render when we're returning plaintext response.

emileten: I see what you're saying, and I agree.

lib/ingestor-api/runtime/src/schemas.py
emileten: Actually, I realized today in another deployment that the `Optional` flag actually means you're assigning `None` when not providing a value explicitly, which can create problems, so we shouldn't merge this as is.

sharkinsspatial:
With respect to the exception response formatting, I'd tend to agree with @alukach here.  In this case, users are not directly parsing the `POST` response body so I think maintaining a focus on single line returns with a status code that is easily parsable from the logs is the best approach.  

With the respect to the `item_assets`, I'd recommend the following
- Let's make a separate PR which implements a feature flag configuration to make this validation configurable.
- When we complete https://github.com/developmentseed/cdk-pgstac/issues/4 we had discussed the option of making validation code pluggable so that any consumer application could specify their own validation module that would be injected into the `cdk-pgstac` construct when the `ingestor` handler is created.  This could be a good first use case for that functionality.

As an aside, can we try to limit the scope of PR's to a functional change.  As an example, the title of this PR is `feat(ingestor): validation exception handler` while the `item_assets` change is related (as I assume this was the validation error you were hitting and wanted better messages about 😄 ) it makes a bit harder to review and track when we mix functional changes in a single PR.

emileten:
> As an aside, can we try to limit the scope of PR's to a functional change. As an example, the title of this PR is feat(ingestor): validation exception handler while the item_assets change is related (as I assume this was the validation error you were hitting and wanted better messages about 😄 ) it makes a bit harder to review and track when we mix functional changes in a single PR.

Yes I agree, but I was hesitant between doing what you say -- split commits based on functionality -- or keep them together to avoid having too many automatic releases happening.

sharkinsspatial:
@emileten I hear your pain on the automated releases.  Thus far I feel like `semantic-release-bot` has been a bit to much magic and makes the release process too opaque.  If @alukach agrees I think we should consider removing it and moving to a manual tag driven release model.

alukach:
I agree that better-scoped PRs are a good thing.  Regarding the future of Semantic Release bot, let's work through that on #37

PR #42, opened 2023-05-22
feat(titiler-pgstac-api): add titiler-pgstac endpoint

emileten:
Solves https://github.com/developmentseed/cdk-pgstac/issues/40 by adding a 'titiler-pgstac' construct, which mirrors the STAC API construct : I essentially 
- copy pasted the lib/stac-api folder content into a new lib/titiler-pgstac-api folder
- removed the `app.py` code and simply wrote `from titiler.pgstac.main import app` in the `handler.py`
- adapted the requirements and `config.py` (`titiler.pgstac.main` requires the postgres credentials to be set as environment variables)


emileten:
I tried to deploy using this, and unfortunately the size of the unzipped package is too large to be used for a lambda function, so I have to use a Dockerfile instead, just like it's done [here](https://github.com/NASA-IMPACT/impact-tiler) or [here](https://github.com/developmentseed/eoAPI/blob/main/dockerfiles/Dockerfile.raster) (any thoughts @vincentsarago ? Thanks !)

Just pushed the Dockerfile, that works, but need to change the CDK code and test it, so I changed the title of this PR to `Draft: *`.

vincentsarago:
@emileten, Dockerimage in Lambda should only be use if absolutely no choice (because it's much slower, at least for dynamic tiling) 

> I think it's because titiler.pgstac has many dependencies

I've deployed titiler.pgstac in lambda before so I doubt this is the case! 


Oh, I misread the comments at first. Yes using simple `requirements.txt` won't work with CDK. Using custom dockerfile **to build** the package is the way to go.

as you can see in the dockerfile we do https://github.com/developmentseed/cdk-pgstac/pull/42/files#diff-6b803bce2d18a876f97ed5ee61d6137408b36add96327274f990ff18d81430a4R11-R16 which is `cleaning` some of the dependencies (already in AWS lambda env). in eoAPI, we also specify `--no-binary pydantic` (https://github.com/developmentseed/eoAPI/blob/main/infrastructure/aws/dockerfiles/Dockerfile.raster#LL9C88-L9C108) which can also reduce the package size a bit 

👍

emileten:
Great that matches with my understanding ! Thank you!

emileten:
Any chance someone takes a look in the following days? 

@sharkinsspatial ? or anyone else 😄

lib/titiler-pgstac-api/index.ts
```diff
@@ -0,0 +1,113 @@
...
+          "PGSTAC_SECRET_ARN": props.dbSecret.secretArn,
+      }
+    
+      
+      const handler = new lambda.Function(this, "lambda", {
+        handler: "handler.handler",
+        runtime: lambda.Runtime.PYTHON_3_8,
+        code: lambda.Code.fromDockerBuild(__dirname, {
+          file: "runtime/Dockerfile",
+          buildArgs: { PYTHON_VERSION: '3.8' },
```
vincentsarago, line 44: FYI: Python 3.10 is available now and might be a better default

vincentsarago (approved):
LGTM but I'm not familiar enough with cdk-pgstac that you might want to wait for @alukach or @sharkinsspatial approval 🤷

lib/titiler-pgstac-api/runtime/src/.env
```diff
@@ -0,0 +1 @@
+PGSTAC_SECRET_ARN="arn:aws:secretsmanager:us-west-2:916098889494:secret:pgstac/bootstrap-pgstac-instance/093a89d4-2hB9wY"
```
alukach, line 1: Should this be committed?  Will it be available to other accounts?

lib/titiler-pgstac-api/index.ts
alukach:
```suggestion
          buildArgs: { PYTHON_VERSION: '3.10' },
```

lib/titiler-pgstac-api/runtime/src/config.py
```diff
@@ -0,0 +1,73 @@
...
+
+        params = dict(
+            postgres_host=secret["host"],
+            postgres_dbname=secret["dbname"],
+            postgres_user=secret["username"],
+            postgres_pass=secret["password"],
+            postgres_port=secret["port"],
+        )
+
+        [set_var(k, v) for k, v in params.items()]
```
alukach, line 68: Beyond being able to simplify this, I think it's a good call to _not_ print the key/values to stdout for security reasons (eg leaking these credentials to Github Actions workflow logs)

```suggestion
        os.environ.update({
            "postgres_host": secret["host"],
            "postgres_dbname": secret["dbname"],
            "postgres_user": secret["username"],
            "postgres_pass": secret["password"],
            "postgres_port": str(secret["port"]),
        })            
```

lib/titiler-pgstac-api/index.ts
```diff
@@ -0,0 +1,113 @@
...
+          "DB_MAX_CONN_SIZE": "1",
+          "PGSTAC_SECRET_ARN": props.dbSecret.secretArn,
+      }
+    
+      
+      const handler = new lambda.Function(this, "lambda", {
+        handler: "handler.handler",
+        runtime: lambda.Runtime.PYTHON_3_8,
+        code: lambda.Code.fromDockerBuild(__dirname, {
+          file: "runtime/Dockerfile",
```
alukach, line 43: To ensure that I understand correctly, if we create assets bundle but don't manually go through at trim it down with the `find ... | xargs rm` commands within the `Dockerfile`, the assets zip will be too large?

lib/titiler-pgstac-api/runtime/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+# Reduce package size and remove useless files
+RUN cd /asset && find . -type f -name '*.pyc' | while read f; do n=$(echo $f | sed 's/__pycache__\///' | sed 's/.cpython-[0-9]*//'); cp $f $n; done;
+RUN cd /asset && find . -type d -a -name '__pycache__' -print0 | xargs -0 rm -rf
+RUN cd /asset && find . -type f -a -name '*.py' -print0 | xargs -0 rm -f
+RUN find /asset -type d -a -name 'tests' -print0 | xargs -0 rm -rf
+RUN rm -rdf /asset/numpy/doc/ /asset/boto3* /asset/botocore* /asset/bin /asset/geos_license /asset/Misc
+
+COPY runtime/src/*.py /asset
+
+CMD ["echo", "hello world"]
```
alukach, line 20:
```suggestion
```

lib/titiler-pgstac-api/runtime/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+
+COPY runtime/requirements.txt requirements.txt
+RUN python -m pip install -r requirements.txt "mangum>=0.14,<0.15" -t /asset
+
+# Reduce package size and remove useless files
+RUN cd /asset && find . -type f -name '*.pyc' | while read f; do n=$(echo $f | sed 's/__pycache__\///' | sed 's/.cpython-[0-9]*//'); cp $f $n; done;
+RUN cd /asset && find . -type d -a -name '__pycache__' -print0 | xargs -0 rm -rf
+RUN cd /asset && find . -type f -a -name '*.py' -print0 | xargs -0 rm -f
+RUN find /asset -type d -a -name 'tests' -print0 | xargs -0 rm -rf
+RUN rm -rdf /asset/numpy/doc/ /asset/boto3* /asset/botocore* /asset/bin /asset/geos_license /asset/Misc
```
alukach, line 16: Opportunity to simplify:

```suggestion
RUN find /asset -type f -name '*.pyc' | while read f; do n=$(echo $f | sed 's/__pycache__\///' | sed 's/.cpython-[0-9]*//'); cp $f $n; done;
RUN find /asset -type d -a -name '__pycache__' -print0 | xargs -0 rm -rf
RUN find /asset -type f -a -name '*.py' -print0 | xargs -0 rm -f
RUN find /asset -type d -a -name 'tests' -print0 | xargs -0 rm -rf
RUN rm -rdf /asset/numpy/doc/ /asset/boto3* /asset/botocore* /asset/bin /asset/geos_license /asset/Misc
```

lib/titiler-pgstac-api/runtime/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+WORKDIR /tmp
+RUN python -m pip install pip -U
+
+COPY runtime/requirements.txt requirements.txt
+RUN python -m pip install -r requirements.txt "mangum>=0.14,<0.15" -t /asset
+
+# Reduce package size and remove useless files
+RUN cd /asset && find . -type f -name '*.pyc' | while read f; do n=$(echo $f | sed 's/__pycache__\///' | sed 's/.cpython-[0-9]*//'); cp $f $n; done;
+RUN cd /asset && find . -type d -a -name '__pycache__' -print0 | xargs -0 rm -rf
+RUN cd /asset && find . -type f -a -name '*.py' -print0 | xargs -0 rm -f
```
alukach, line 14: Am I missing something, or does this drop all of the Python files in our assets dir?

alukach (changes requested):
Strongly recommended changes:

1. `git rm lib/titiler-pgstac-api/runtime/src/.env`
1. Don't write the pgstac secret values to stdout.


Otherwise, I think this looks good.  Left some other comments but those are mostly focused on upping my understanding on how this work.

lib/titiler-pgstac-api/runtime/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+FROM --platform=linux/amd64 public.ecr.aws/lambda/python:${PYTHON_VERSION}
+
+WORKDIR /tmp
+RUN python -m pip install pip -U
+
+COPY runtime/requirements.txt requirements.txt
+RUN python -m pip install -r requirements.txt "mangum>=0.14,<0.15" -t /asset
+
+# Reduce package size and remove useless files
+RUN cd /asset && find . -type f -name '*.pyc' | while read f; do n=$(echo $f | sed 's/__pycache__\///' | sed 's/.cpython-[0-9]*//'); cp $f $n; done;
```
alukach, line 12: This line kind of blows my mind. So, we're finding all `*.pyc` files, reading through them line by line, then copying the `.pyc` file to the path mentioned within the file after removing `__pycache__` and/or `.cpython*` from the path?

What does this do?

vincentsarago: @alukach 

> So, we're finding all *.pyc files, reading through them line by line, then copying the .pyc file to the path mentioned within the file after removing __pycache__ and/or .cpython*

Yep, we get rid of the non compiled `.py` files that are useless because we have the `.pyc` files. This wins maybe 10ms when we start the handler 😄

vincentsarago: This is just a small optimization I do on all my lambdas

lib/titiler-pgstac-api/runtime/Dockerfile
vincentsarago: I think I tried that in the past but maybe I was getting some weird result. Should def try to build the package before changing ☝️

lib/titiler-pgstac-api/runtime/Dockerfile
vincentsarago: we need to keep this line or CDK will complain when building the lambda package.

lib/titiler-pgstac-api/runtime/Dockerfile
alukach: So do these compiled files get created from the `pip install` command? Is there any value to running `python -m compileall .` before this?

Btw, doing some research on this, I found [this SO answer](https://stackoverflow.com/a/53079956/728583) which linked to [this article] (https://blog.mapbox.com/aws-lambda-python-magic-e0f6a407ffc6). I was about to link it until I realized that **you wrote the article.** 😂

lib/titiler-pgstac-api/runtime/src/.env
emileten: Ugh 🤯 I did not mean to commit this.

emileten: I removed this file (using `git rm` like you suggested), however, I don't think it entirely erases it from the history, like it still shows up in this PR. 

That said, it doesn't sound like an 'unsafe secret to leak', since it's just an ARN ? I.e it doesn't sound like I need to go and redeploy so that the ARN changes. What do you think @alukach ?

lib/titiler-pgstac-api/runtime/Dockerfile
emileten: I will follow @vincentsarago's comment and resolve without the suggested change, since I took this Dockerfile (almost) straight from [eoAPI](https://github.com/developmentseed/eoAPI/blob/main/infrastructure/aws/dockerfiles/Dockerfile.raster). The same command is run [here](https://github.com/NASA-IMPACT/impact-tiler/blob/master/infrastructure/lambda/Dockerfile).

lib/titiler-pgstac-api/runtime/src/config.py
emileten: That's a piece of code I used for testing and left it there... Thanks @alukach for flagging.

lib/titiler-pgstac-api/runtime/Dockerfile
vincentsarago: > So do these compiled files get created from the pip install command? Is there any value to running python -m compileall . before this?

Yes they are created on `pip install` command

vincentsarago: > Btw, doing some research on this, I found [this SO answer](https://stackoverflow.com/a/53079956/728583) which linked to [this article](https://blog.mapbox.com/aws-lambda-python-magic-e0f6a407ffc6). I was about to link it until I realized that you wrote the article. 😂

Old days ;-)

lib/titiler-pgstac-api/runtime/Dockerfile
emileten: a side note, the first of these commands (the one for the `*.pyc` files) takes quite a long time, around 90 secs.

emileten:
@alukach I added most of your suggestions -- good to merge ?

lib/titiler-pgstac-api/runtime/src/handler.py
```diff
@@ -0,0 +1,11 @@
+"""
+Handler for AWS Lambda.
+"""
+
+from mangum import Mangum
+from config import ApiSettings
+from titiler.pgstac.main import app
+
+settings = ApiSettings()
+settings.set_postgres_settings()
```
vincentsarago, line 10: This is kinda useless, we just need to update the environment variables for the `PostgresSettings`
https://github.com/stac-utils/titiler-pgstac/blob/main/titiler/pgstac/settings.py#L36-L51

```python
"""
Handler for AWS Lambda.
"""

import os
from mangum import Mangum
from config import get_secret_dict
from titiler.pgstac.main import app

pgstac_secret_arn = os.environ["PGSTAC_SECRET_ARN"]

secret = get_secret_dict(pgstac_secret_arn)
os.environ.update(
    {
        "postgres_host": secret["host"],
        "postgres_dbname": secret["dbname"],
        "postgres_user": secret["username"],
        "postgres_pass": secret["password"],
        "postgres_port": str(secret["port"]),
    }
)

handler = Mangum(app)
```

vincentsarago: I find this a bit `hacky`, I think we will be better served with a proper `custom` application! 

```python

import os
from mangum import Mangum

from titiler.pgstac.main import app
from titiler.pgstac.settings import PostgresSettings

import boto3


def get_secret_dict(secret_name: str):
    """Retrieve secrets from AWS Secrets Manager
    Args:
        secret_name (str): name of aws secrets manager secret containing database connection secrets
        profile_name (str, optional): optional name of aws profile for use in debugger only
    Returns:
        secrets (dict): decrypted secrets in dict
    """

    # Create a Secrets Manager client
    session = boto3.session.Session()
    client = session.client(service_name="secretsmanager")

    get_secret_value_response = client.get_secret_value(SecretId=secret_name)

    if "SecretString" in get_secret_value_response:
        return json.loads(get_secret_value_response["SecretString"])
    else:
        return json.loads(base64.b64decode(get_secret_value_response["SecretBinary"]))


@app.on_event("startup")
async def startup_event() -> None:
    """Connect to database on startup."""
    
    pgstac_secret_arn = os.environ["PGSTAC_SECRET_ARN"]
    secret = get_secret_dict(pgstac_secret_arn)
    
    await connect_to_db(
        app,
        settings=PostgresSettings(
            postgres_user: secret["username"],
            postgres_pass: secret["password"],
            postgres_host: secret["host"],
            postgres_port: str(secret["port"]),
            postgres_dbname: secret["dbname"],
        )
    )


handler = Mangum(app)
```

vincentsarago: One more comment! any reason why not set those variable in the environment directly? https://github.com/developmentseed/eoAPI/blob/main/infrastructure/aws/cdk/app.py#L196-L212

lib/titiler-pgstac-api/runtime/src/.env
alukach: I agree, I don't typically think of ARNs as things we need to be worried about leaking (others disagree with this view).  I think a simple `git rm` is good enough.

alukach:
> @alukach I added most of your suggestions -- good to merge ?

@emileten I agree with Vincent that injecting secrets as env variables at the CDK level would probably be simpler than doing so in the runtime of the code, but I'm good either way and think you can merge when you feel this is ready.

lib/titiler-pgstac-api/runtime/src/handler.py
vincentsarago: 🤔 the `build_app` approach seems interesting

alukach (approved):
@emileten Realized I was still marked as "requested changes".  I'm good with merging if you are.

alukach (approved):
@emileten Realized I was still marked as "requested changes".  I'm good with merging if you are.

emileten:
Thanks @alukach ! Should be -- just wanted to check the last comment I haven't read above tomorrow.

lib/titiler-pgstac-api/runtime/src/handler.py
emileten: I simplified following your first comment @vincentsarago.

I think what you describe in your second code suggestion would be ideal, but since it's not critical and would require me re-testing this in deployment before merging, I will stick with the current approach (set up the env in handler.py) and we can always improve things later.

PR #45, opened 2023-06-11
fix(bootstrapper): bootstrapper should use httpx

abarciauskas-bgse:
Addresses https://github.com/developmentseed/cdk-pgstac/issues/44 by using httpx instead of requests in bootstrapper.

I tested this with cdk-pgstac deployment I am using for EODC (see https://github.com/developmentseed/tile-benchmarking/tree/feat/deploy-pgstac) where I made these changes in the installed node_module for cdk-pgstac and deployed.


vincentsarago (approved):
❤️

PR #62, opened 2023-08-04
feat: tipg-api

### LINKED ISSUE #51: Add a vector application construct
Following a slack thread, if this repository is to contain all the constructs for CDK based deployment of all [eoAPI](https://github.com/developmentseed/eoAPI) services, we want to also have a construct and the FastAPI components for a vector application.

It would be pretty simple and quick :

```[tasklist]
### Tasks
- [ ] create a `/lib/tipg-api` folder
- [ ] take the construct from this [chunk of eoapi python code](https://github.com/developmentseed/eoAPI/blob/cad43784773dd3241602961e40567c24c40595ed/infrastructure/aws/cdk/app.py#L358C1-L399C55), convert it to TS and put it in the folder
- [ ] take the fastAPI components from this chunk of eoapi python code and put them in a `runtime` subfolder, that is going to be referenced by the TS construct
```

vincentsarago:
closes #51 

## Todo 

- [x] Add docs (readme)


lib/tipg-api/index.ts
```diff
@@ -0,0 +1,112 @@
...
+        environment: {
+          PGSTAC_SECRET_ARN: props.dbSecret.secretArn,
+          DB_MIN_CONN_SIZE: "1",
+          DB_MAX_CONN_SIZE: "1",
+          ...props.apiEnv,
+        },
+        vpc: props.vpc,
+        vpcSubnets: props.subnetSelection,
+        allowPublicSubnet: true,
+        memorySize: 1024,
```
vincentsarago, line 44: IMO this should be configurable

emileten:
@vincentsarago is the only TODO left to add a readme ? It looks like the rest is done to me, but I'm not 100% sure because I haven't worked with tipg yet.

vincentsarago:
@emileten yes, I'll try to pickup this PR tomorrow but if you want to have a look and check that I haven't done anything stupid feel free 🙏

lib/tipg-api/index.ts
emileten: It makes sense to me as well @vincentsarago.

Should we also make [the titiler-pgstac one configurable ?](https://github.com/developmentseed/eoapi-cdk/blob/daec1e01c4d57f9806b5e34b4d1424fc56146106/lib/titiler-pgstac-api/index.ts#L51) then ?

vincentsarago: yes, we can make all of them configurable in another PR!

emileten:
@vincentsarago I integrated this to our maap-eoapi stack, and it deployed successfully, however a get request to the resulting `tipg` HTTP API returns me `Internal Server Error`. Here is the lambda log :  

```
[ERROR] Runtime.ImportModuleError: Unable to import module 'src.handler': No module named 'utils'
```

lib/tipg-api/index.ts
```diff
@@ -0,0 +1,112 @@
...
+  import { Construct } from "constructs";
+
+  export class TiPgApiLambda extends Construct {
+    readonly url: string;
+    public tiPgLambdaFunction: PythonFunction;
+
+    constructor(scope: Construct, id: string, props: TiPgApiLambdaProps) {
+      super(scope, id);
+
+      const apiCode = props.apiCode || {
```
emileten, line 25: I was using the default api code. These few lines could be the culprits ? If I compare to titiler-pgstac the set up is different. We use 'assets from docker build' for titiler-pgstac, and I just checked, in the Dockerfile we explicitly transfer _all_ the python files (including `utils.py`) to the Docker filesystem. Are we sure this happens here ?

vincentsarago: @emileten I'm not 💯 sure I follow everything, can you point me toward the code you are using. 

For this `apiCode` I think I copy/pasted from the STAC API code...

emileten: I need to push the changes to the wrapper I'm using @vincentsarago, haven't done it yet. I might be doing something wrong in it, but I'm just calling the new construct of this PR without passing any value to apiCode, so the default should be used.

I'll look for the cause of this problem tomorrow !

emileten: [solved here.](https://github.com/developmentseed/eoapi-cdk/pull/62/commits/26ad3134403e79684788cbf54709d60177d261ab)

it's the same problem that was caused by https://github.com/developmentseed/eoapi-cdk/pull/61 but there it was with the tiler.

Here, the problem was : we do `from tipg.main import app` at the top of the handler. Doing this import [triggers a postgres settings creation](https://github.com/developmentseed/tipg/blob/1fc3b73c17da45826dc821414583059459e7b1b1/tipg/main.py#L28), but the postgres environment variables are _not_ yet loaded into the environment, it was happening [_after_ all imports](https://github.com/developmentseed/eoapi-cdk/blob/83304a61bcfacba3ecc292bb557a4c14fee494a5/lib/tipg-api/runtime/src/handler.py#L21). There are only two ways around this : 

1. load the postgres env vars before the imports (we don't want to have them in the lambda configuration directly, because it exposes secrets). That's [what I did here.](https://github.com/developmentseed/eoapi-cdk/pull/62/commits/26ad3134403e79684788cbf54709d60177d261ab) 
2. _don't_ rely on `from tipg.main import app` and have a custom app instead. [That's what @vincentsarago was suggesting we do here](https://github.com/developmentseed/eoapi-cdk/pull/42#discussion_r1212805079), and that's what [we do for stac-fastapi](https://github.com/developmentseed/eoapi-cdk/blob/main/lib/stac-api/runtime/src/app.py) (however I'm not sure the stac-fastapi provide an app layer that's reusable like tipg or titiler-pgstac). 

Back when I added the titiler pgstac construct here I decided to reuse the standard app layer from the titiler package (titiler.main.app) nbecause I saw no reason to duplicate code. 

Anyways, just like for the tiler bug, I chose (1) to solve this bug.

emileten:
[solved here.](https://github.com/developmentseed/eoapi-cdk/pull/62/commits/26ad3134403e79684788cbf54709d60177d261ab)

it's the same problem that was caused by https://github.com/developmentseed/eoapi-cdk/pull/61 but there it was with the tiler.

Here, the problem was : we do `from tipg.main import app` at the top of the handler. Doing this import [triggers a postgres settings creation](https://github.com/developmentseed/tipg/blob/1fc3b73c17da45826dc821414583059459e7b1b1/tipg/main.py#L28), but the postgres environment variables are _not_ yet loaded into the environment, it was happening [_after_ all imports](https://github.com/developmentseed/eoapi-cdk/blob/83304a61bcfacba3ecc292bb557a4c14fee494a5/lib/tipg-api/runtime/src/handler.py#L21). There are only two ways around this : 

1. load the postgres env vars before the imports (we don't want to have them in the lambda configuration directly, because it exposes secrets). That's [what I did here.](https://github.com/developmentseed/eoapi-cdk/pull/62/commits/26ad3134403e79684788cbf54709d60177d261ab) 
2. _don't_ rely on `from tipg.main import app` and have a custom app instead. [That's what @vincentsarago was suggesting we do here](https://github.com/developmentseed/eoapi-cdk/pull/42#discussion_r1212805079), and that's what [we do for stac-fastapi](https://github.com/developmentseed/eoapi-cdk/blob/main/lib/stac-api/runtime/src/app.py) (however I'm not sure the stac-fastapi provide an app layer that's reusable like tipg or titiler-pgstac). 

Back when I added the titiler pgstac construct here I decided to reuse the standard app layer from the titiler package (titiler.main.app) nbecause I saw no reason to duplicate code. 

Anyways, just like for the tiler bug, I chose (1) to solve this bug.

emileten (approved):
@vincentsarago I think it's now good to go (see above), feel free to merge. Two things to note : 

- Could you point me to an example running tipg api ? I am not familiar with the technology and I would like to have a look. Is the eoapi dev a good place to look at ? 
- Do you have examples of inserting data into the pgstac database that tipg can read ? I tried deploying this connected to the MAAP pgstac database but we only have STAC records there, so nothing that tipg could read, if I understand well. 
- We should add an option for a tipg CDN later, just like we did [here for the stac-api, the stac-ingestor, and titiler-pgstac. ](https://github.com/developmentseed/eoapi-cdk/pull/63)

vincentsarago:
> Could you point me to an example running tipg api ? I am not familiar with the technology and I would like to have a look. Is the eoapi dev a good place to look at ?

Yes, https://vector.eoapi.dev

> Do you have examples of inserting data into the pgstac database that tipg can read ?

Not really, tipg is not meant to render data from pgstac but maybe ingested in other schemas. We can render `pgstac` tables directly but the schema is not optimized for this use case. The idea of tipg is to be able to store vector dataset that could be used in a frontend application in addition to STAC metadata and Raster tiles

> We should add an option for a tipg custom domain name later, just like we did https://github.com/developmentseed/eoapi-cdk/pull/63

I think we should do it in the PR as well 🙏 


Thank @emileten for this nice review and fixes 🙌

emileten:
> I think we should do it in the PR as well 🙏

Oh yes let's do it here. 

> Not really, tipg is not meant to render data from pgstac but maybe ingested in other schemas. We can render pgstac > tables directly but the schema is not optimized for this use case. The idea of tipg is to be able to store vector dataset > that could be used in a frontend application in addition to STAC metadata and Raster tiles

Ok maybe I need to rephrase my question @vincentsarago and it also shows the limit of my knowledge about databases. So we're connecting `tipg` to the same RDS instance as `titiler-pgstac` and `stac-api`, [here](https://github.com/developmentseed/eoAPI/blob/8e97d74e08ae05f0da8a36656bc247579a35a81d/infrastructure/aws/cdk/app.py#L372) for example. Is `tipg` going to read from another database in the same instance ? Do you have examples of ingesting data into an RDS instance that is then read by `tipg`?

PR #63, opened 2023-08-16
feat: custom domain names for apis

jjfrench:
- adds the ability for wrappers to pass custom domain names to apigateway constructs
- fixes a titiler-pgstac application bug introduced in https://github.com/developmentseed/eoapi-cdk/pull/61

example implementation: https://github.com/MAAP-Project/maap-eoapi/pull/15

### DIFF
--- a/lib/ingestor-api/index.ts
+++ b/lib/ingestor-api/index.ts
@@ -191,6 +191,7 @@ export class StacIngestor extends Construct {
     stage: string;
     policy?: iam.PolicyDocument;
     endpointConfiguration?: apigateway.EndpointConfiguration;
+    ingestorDomainNameOptions?: apigateway.DomainNameOptions;
   }): apigateway.LambdaRestApi {
     return new apigateway.LambdaRestApi(
       this,
@@ -205,6 +206,11 @@ export class StacIngestor extends Construct {
 
         endpointConfiguration: props.endpointConfiguration,
         policy: props.policy,
+
+        domainName:  props.ingestorDomainNameOptions ? {
+          domainName: props.ingestorDomainNameOptions.domainName,
+          certificate: props.ingestorDomainNameOptions.certificate,
+        } : undefined,
       }
     );
   }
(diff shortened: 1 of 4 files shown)


lib/ingestor-api/index.ts
```diff
@@ -62,6 +62,7 @@ export class StacIngestor extends Construct {
       stage: props.stage,
       endpointConfiguration: props.apiEndpointConfiguration,
       policy: props.apiPolicy,
+      ingestorDomainNameOptions: props.ingestorDomainNameOptions,
```
emileten, line 65: We had missed this which meant we weren't _actually_ using the domain name option for the ingestor. I added it. I tested in https://github.com/MAAP-Project/maap-eoapi/pull/15 that it works.

lib/titiler-pgstac-api/runtime/src/handler.py
```diff
@@ -22,6 +20,9 @@
     }
 )
 
+from titiler.pgstac.main import app  # noqa: E402
```
emileten, line 23: Here I am fixing a bug introduced in https://github.com/developmentseed/eoapi-cdk/pull/61.

When you import from `titiler.pgstac.main`, you trigger the creation of the postgres settings https://github.com/stac-utils/titiler-pgstac/blob/50e37883ffe12cc739467f25d370c00672fe9070/titiler/pgstac/main.py#L42. So, before doing that import, you need to make sure all the variables you need for these settings are in the environment. 

I swapped the import and the postgres env setup, and turned off the flake8 E402 rule ("module level import not at top of file") for these imports.

emileten (approved):
I pushed two commits to fix two bugs I described in comments. LGTM now.

alukach (approved):

lib/titiler-pgstac-api/runtime/src/handler.py
vincentsarago: oh Sorry about this 🙏

vincentsarago (approved):

PR #64, opened 2023-08-22
feat: add STAC browser option

emileten:
Add a construct defining an S3 deployment with a Radiant Earth static STAC browser. ~~The user needs to compile the website code first locally, and provide the path to the compiled code directory to the construct parameters~~  _The build process is automated, see comments below_. By default the bucket is deployed with no endpoint or no particular permissions ; however : 

- a `cloudFrontDistributionArn` option is provided in the construct props. If provided, this ARN is granted read access to the bucket. This is useful to enable HTTPS.
- a  `websiteIndexDocument` option is provided, to enable static website hosting. 
- the `Bucket` and `BucketDeployment` objects are accessible as fields of the construct, in case clients need to modify them (e.g. change the permissions on the bucket).


vincentsarago:
> The user needs to compile the website code first locally, and provide the path to the compiled code directory to the construct parameters. 

Ah 😭, I was hopping there was a way to automate this

emileten:
> > The user needs to compile the website code first locally, and provide the path to the compiled code directory to the construct parameters.
> 
> Ah 😭, I was hopping there was a way to automate this

@vincentsarago it's possible -- back when I wrote this I just didn't want to spend time to try and my idea would have lead to something that's not very clean. But here it is if you want to take a look -- my last commit. The idea is simply to run _within_ the construct, the following commands required to build the app : 

```
- npm install
- npm run build
```

These commands are executed as external processes. This commit allows clients to avoid having to do this.

vincentsarago (approved):
Things look great as far as I can understand 😅

lib/stac-browser/index.ts
```diff
@@ -0,0 +1,113 @@
...
+                    resources: [this.bucket.arnForObjects('*')],
+                    conditions: {
+                        'StringEquals': {
+                            'aws:SourceArn': props.cloudFrontDistributionArn
+                        }
+                    }
+                }));
+        
+
+        this.bucketDeployment = new s3_deployment.BucketDeployment(this, 'BucketDeployment', {
```
jjfrench, line 38: Could I choose an existing bucket?

lib/stac-browser/index.ts
```diff
@@ -0,0 +1,113 @@
...
+
+        // if `cloneDirectory` exists, delete it
+        if (fs.existsSync(cloneDirectory)) {
+            console.log(`${cloneDirectory} already exists, deleting...`)
+            execSync(`rm -rf ${cloneDirectory}`);
+        }
+
+        // Clone the repo
+        console.log(`Cloning ${githubRepoUrl} into ${cloneDirectory}`)
+        execSync(`git clone ${githubRepoUrl} ${cloneDirectory}`);
```
jjfrench, line 64: Would checking the existence of the cloneDirectory and then fetching the tags work instead of deleting? or is there another purpose to deleting the cloned repository every deploy?

lib/stac-browser/index.ts
emileten: Added an option to use an existing bucket, and added a note that `cloudFrontDistributionArn` has no effect in that case, and that the identity doing the deployment should have appropriate permissions.

emileten:
@jjfrench in case you didn't see (not sure how well that feature works), I re-requested a review (addressed your comments). We should be good to merge, however I wanted to ask, did you face any issue when trying to deploy this _as part_ of maap-eoapi ?

jjfrench (approved):

PR #66, opened 2023-09-01
feat: custom runtimes option for titiler and ingestor

emileten:
Closes https://github.com/developmentseed/eoapi-cdk/issuhttps://github.com/developmentseed/eoapi-cdk/pullses/55. Adds an option for custom python lambda runtime code for : 

- the titiler construct
- the ingestor construct

For the ingestor, note that I added _two_ options (`apiCode` and `ingestorCode`), since we're creating _two_ lambdas : 

- one for the API in itself that puts the ingestion requests in a queue
- one that gets the queued ingestion requests and bulk-inserts them in the database. 

Defaults are what's in the `runtime` folders here.


lib/titiler-pgstac-api/index.ts
```diff
@@ -40,15 +40,15 @@ import {
...
         }),
         timeout: Duration.seconds(30),
         vpc: props.vpc,
         vpcSubnets: props.subnetSelection,
         allowPublicSubnet: true,
-        memorySize: 3008,
+        memorySize: props.titilerLambdaMemorySize ?? 3008,
         logRetention: aws_logs.RetentionDays.ONE_WEEK,
         environment: titilerPgstacEnv,
       });
```
vincentsarago, line 54: Can we switch to `PythonFunction` like for the other construct?

emileten: Wasn't the purpose of using a docker build to be able to [trim a little bit the package through the dockerfile ?](https://github.com/developmentseed/eoapi-cdk/blob/a94cf94cd8c0d7270cacb678f545a7bc14a54af9/lib/titiler-pgstac-api/runtime/Dockerfile#L12) 

If we move to `PythonFunction`, [can we do the same for the database `bootstrapper`? ](https://github.com/developmentseed/eoapi-cdk/blob/a94cf94cd8c0d7270cacb678f545a7bc14a54af9/lib/bootstrapper/index.ts#L32)

emileten: > As is, it doesn't work with PythonFunction because the package is too big.

Looks like `boto3` (which requires `botocore`) is the main cuplrit, it's taking a third of the size. It looks like it's supposed to be included in the lambda python runtime, so we shouldn't need to install it. I am going to remove it from the main requirements to the dev requirements and see if I can deploy this as a `PythonFunction` with that trick.

emileten: Ok that works in my last commit

vincentsarago: Looking at the CdK docs I though we could still use custom docker build ... but maybe not! That will be definitely a blocker if not

emileten:
@vincentsarago we're good, I switched to `PythonFunction` and just removed the `boto3` requirement that we don't need because it's already in the environment provided by AWS. 

Anything else ?

lib/titiler-pgstac-api/runtime/Dockerfile
```diff
@@ -1,20 +0,0 @@
-ARG PYTHON_VERSION
```
vincentsarago, line 1: @emileten we should try to keep this and use method described in https://docs.aws.amazon.com/cdk/api/v2/python/aws_cdk.aws_lambda_python_alpha/README.html#custom-bundling

emileten: solution for this was to make things more configurable, in particular letting users pass `BundlingOptions`, however for the default we're not keeping the docker image. 

`BundlingOptions` allows to provide a custom bundling docker image, or docker commands to run after bundling (via `commandHooks`) -- which could be useful for trimming the dependencies like we were doing in this Dockerfile -- etc.. https://docs.aws.amazon.com/cdk/api/v2/docs/@aws-cdk_aws-lambda-python-alpha.BundlingOptions.html

emileten:
Could you approve if it looks good 🙏

vincentsarago (approved):

PR #67, opened 2023-09-05
feat: add integration test

### LINKED ISSUE #13: Add integration tests.
Initial deployment of this construct revealed several small CDK and configuration related issues that made the resulting infrastructure fail when used in a stack.  ref https://github.com/developmentseed/cdk-pgstac/commit/be40af70326f04d7ed9583569ad49f12da416d6f and ref https://github.com/developmentseed/cdk-pgstac/pull/8

We should build an integration test suite which deploys an ephemeral stack using the construct, it should

- Create a collection via the Ingestor API.
- Ingest an item via the Ingestor API.
- Verify the creation of the item via the STAC API endpoint.

Given the expense (in deployment time and resources) these integration tests should likely only run against new release tags.
```[tasklist]
### Tasks
- [ ] https://github.com/developmentseed/cdk-pgstac/issues/47
- [x] Create a test suite : https://github.com/developmentseed/eoapi-tests
- [x] Create a github action that runs the demo deployment, runs the tests on it, and tears down the deployment
```

emileten:
Closes #13


vincentsarago:
@emileten I'm going to close this PR because there seems to be an issue with the commit history. I'll restart a new PR with only the `deploy.yaml` file

PR #69, opened 2023-09-05
feat: add integration tests

### LINKED ISSUE #13: Add integration tests.

vincentsarago:
closes #13 

second take at #67


.github/workflows/deploy.yaml
```diff
@@ -0,0 +1,78 @@
...
+  deploy-and-integration-test:
+    runs-on: ubuntu-latest
+
+    steps:
+
+    - name: Checkout repository with template stack
+      uses: actions/checkout@v3
+      with:
+        repository: developmentseed/eoapi-template
+        path: eoapi-template
```
vincentsarago, line 20: @emileten @sharkinsspatial I feel that having an external repo to host only the `tests` is maybe over complex. 

IMO we should have the integration tests code within the `.github/` directory OR have everything within the eoapi-tests repo (deploy + test) 

We already have a lot of `eoapi-*` repo and this might become confusing

.github/workflows/deploy.yaml
```diff
@@ -0,0 +1,78 @@
...
+        cd tests
+        python -m venv .tests_venv
+        source .tests_venv/bin/activate
+        pip install -e tests
+        pytest eoapi-tests
+        deactivate
+        cd ..
+
+    - name: Tear down the stack
+      run: |
```
vincentsarago, line 73: should have an `always` tag here to make sure we always tear down the stack

.github/workflows/deploy.yaml
vincentsarago: if we're going to re-use this we can add it directly on the root so other project can reuse it.

vincentsarago: ideally it would be cool if we could have a testing module as part of the `eoapi-cdk` library. But this will mean, rewrite everything in typescript instead of python.

maybe @alukach @sharkinsspatial have opinions 😅

emileten: > if we're going to re-use this we can add it directly on the root so other project can reuse it.

root of `eoapi-cdk` works for now ? (the tests are _slightly_ coupled with AWS because of auth)

emileten: Done, I moved the tests here

emileten:
Anybody would like to take another look ? The action is running fine (I ran them manually to remove any bug). Here is a gist of what's happening : 

- Every time a `Distribute` workflow is going to complete successfully (aka every time we `release`)
- The `Deploy & Test Deployment` workflow I define here picks up this release
- Deploys the default `https://github.com/developmentseed/eoapi-template` in the DevSeed AWS account
- Runs integration tests I define here
- _Always_ tears down the stack at the end. 

It takes a while, almost an hour from scratch : 
- CDK synth ~ 6 minutes
- Deployment from scratch ~ 16 minutes
- Tests ~ 1 min
- Stack destruction ~ 30 minutes. 

For another time I'd like to figure out how to cache all the lambda runtime builds but it's a rabbit hole....

.github/workflows/deploy.yaml
vincentsarago: @emileten Q: why do we need `eoapi-template`? IMO we should have everything we need here and not rely on a third party repo.

emileten: 🤔

Here, we define the constructs. If we want to integration-test, we need to instantiate and wrap all these constructs together. That's exactly what `eoapi-template` is.

Of course I don't think we want to copy-paste that wrapper code here.... 

If we don't like having too many moving repos, a solution would be to remove `eoapi-template` altogether and have its content simply in e.g. `eoapi-cdk/example`. 

@vincentsarago

emileten:
@vincentsarago @sharkinsspatial, following-up on a discussion I had with Vincent on Slack. 

#### There is a concern regarding the fact that we are going to run the `deploy and test` workflow using the `eoapi-template` IaC repo.

What's the  concern ? Let's assume that I submit a PR to `eoapi-cdk` with a breaking change, and I don't correspondingly modify the `eoapi-template` repo IaC. If I merge this PR, a release happens, which triggers a deployment using the new release wrapped into the non-updated `eoapi-template` code. This deployment is going to fail. It _would not have_ failed, had I correspondingly updated `eoapi-template`. 

This means that for any PR that triggers a release, we need to make sure we update `eoapi-template` if needed. This 'makes sense', because in this framework, `eoapi-template` became part of the test suite, and any test suite must be updated if breaking changes are merged. 


## Pros

1. `eoapi-template`, which was meant to be an official example, is tested and up to date. 
2. we don't have to write wrapper IaC just for the purpose of tests, instead we reuse the official example. I am worried about writing IaC _just_ for the tests and ending up with a lot of IaC to manage (`eoapi-template`, test IaC, the constructs...).


## Cons

1. `eoapi-template` has to be modified often. 

_To this, I answer "yes but not that much often, because we have to update it only for breaking change releases. I would also answer that we *do want* to update eoapi-template often anyways because it's the official example"_

2. It's hard to remember to update it because it's in a separate repo. 

_To this, I answer  "we'd have to update any IaC we write for the purposes of the deploy-test workflow, whether it's eoAPI template or not. If the difficulty is that it's hard to remember because it's in a separate repo, I suggest that we move eoapi-template to be within eoapi-cdk."_

vincentsarago:
> 1. eoapi-template has to be modified often.
>
> To this, I answer "yes but not that much often, because we have to update it only for breaking change releases. I would also answer that we do want to update eoapi-template often anyways because it's the official example"

eoapi-template will have to be modified on every release because we are using direct version pinning in eoapi-template (https://github.com/developmentseed/eoapi-template/blob/main/requirements.txt#L3)

I personally having circular dependency with external repo is dangerous is will required more work (e.g if someone change something in eoapi-cdk but has no idea what is eoapi-template is, they might open a PR and wait for someone to review, which mean that the reviewer might have to review two PR in two different repo.

I do agree that eoapi-template should be the official demo application but I do not agree that we have to keep 1:1 with the versions. 


> I suggest that we move eoapi-template to be within eoapi-cdk."

Because the purpose of eoapi-template is to be a repository template, I think we should could it separate. IMO it will be fine to just copy/paste what application we have in eoapi-cdk test suite to eoapi-template

emileten:
> Because the purpose of eoapi-template is to be a repository template, I think we should could it separate. IMO it will be > fine to just copy/paste what application we have in eoapi-cdk test suite to eoapi-template 

👍

emileten:
@vincentsarago made the changes, the tests here use their own wrapper CDK code. Looks like I adressed your main requests, anything else ? Once the action succeeded I will turn off the `on: push` and leave only the "on-release" trigger and manual option.

.github/workflows/deploy.yaml
```diff
@@ -0,0 +1,99 @@
...
+      with:
+        node-version: '18'
+      
+    - name: Install deployment environment
+      working-directory: integration_tests/cdk
+      run: |
+        python -m venv .deployment_venv
+        source .deployment_venv/bin/activate
+        pip install -r requirements.txt
+        pip install --upgrade eoapi-cdk # make sure to take the latest version of eoapi-cdk since we're integration-testing the newest release.
```
vincentsarago, line 51: Are we testing the latest available version on pypi or the one in this repo? 

It might be `dangerous` to test the version on pypi because there could be a `lag` from the push to the publication

vincentsarago:
Thanks @emileten 
I'm wondering what's the purpose of doing integration test after pushing packages on npm and pypi. IMO we should do integration tests and then publish.

.github/workflows/deploy.yaml
vincentsarago: > How about setting up this workflow so that it waits (with a timeout) for the latest pypi to match for the latest github tag?

Sure, but then what will it means if the tests fails?

emileten: Ah 🤦‍♂️ doesn't make much sense indeed. We should be doing

build -> integration test -> release

vincentsarago: lets see if @sharkinsspatial or @alukach have opinions

emileten: Could be missing something from the CI logic, but one way to do it could be to add a [`job` named `integration-test` here](https://github.com/developmentseed/eoapi-cdk/blob/a1081a22e034d50f89d248d36844d598d45d5c69/.github/workflows/distribute.yaml#L11) that reuses the `package` job's output. We then add `integration-test` as a dependency to both subsequent `distribute-*` jobs so that we're sure releases happen only if integration-tests ran.

emileten: FYI, to illustrate the idea, I pushed another commit (can always revert). See https://github.com/developmentseed/eoapi-cdk/pull/69/commits/701d85f51a5359cc31d07e37b3e004ea205b9302.

sharkinsspatial: @emileten I would recommend relying on the local python modules from the output of the TS build from the `package` output as you outlined in https://github.com/developmentseed/eoapi-cdk/commit/4ff82c23c0a09f7b0cd5d42308a3f851f9757a87.

sharkinsspatial:
@emileten I concur with @vincentsarago here and I like all the changes made to have the local integration test app self-contained within the repo.

After this PR is merged, we should consider deprecating eoapi-template and point the documentation to the integration test app stack to avoid duplication and needing to maintain both.

sharkinsspatial (approved):

emileten:
Sounds good @sharkinsspatial

integration_tests/tests/eoapi_tests/test_stac_ingestion.py
```diff
@@ -0,0 +1,123 @@
...
+# Test querying collection and verifying inserted collection
+pytest.mark.order(4)
+
+
+def test_query_collection(stac_ingestion_instance, test_collection):
+    response = stac_ingestion_instance.query_collection(test_collection["id"])
+    assert response.status_code in [
+        200,
+        201,
+    ], f"Failed to query the test_collection :\n{response.text}"
```
gadomski, line 61: This would be a good opportunity to validate the collection -- currently, collections are serialized _without_ `exclude_none` or `exclude_unset`: https://github.com/developmentseed/eoapi-cdk/blob/239b0a9b19090a7da7d8790cd23cc70257431f03/lib/ingestor-api/runtime/src/collection.py#L22

This leads to invalid STAC in the database (e.g. `assets` is often set to `null`). We could catch those problems in these integration tests by validating (e.g. with `pystac.Collection.from_dict(response.json()).validate()`).

vincentsarago:
@emileten is there any specific thing to do here?

emileten:
> @emileten is there any specific thing to do here?

@vincentsarago I think it's almost good to merge but I wanted to get this one merged first https://github.com/developmentseed/eoAPI/pull/144

vincentsarago:
@emileten any chance we could get this PR merged?

vincentsarago:
@emileten can we merge this? we're going to clean up a bit all the eoapi-* repo next week (mostly to make eoapi-template an important piece for onboarding new users)

emileten:
@vincentsarago sorry I let this become stale. Resuming work on this one today !

emileten:
Still in progress ; I am having issues finding the builds in the local filesystem of the github runner, to deploy : https://github.com/developmentseed/eoapi-cdk/actions/runs/7962946517/job/21737522652.

emileten:
Making progress on this, now I am trying to figure out why I get a TimeOutError from httpx in the bootstrapper. 

Because this bootstrapper runs fine on my local machine, my guess is that the bootstrapper lambda doesn't have access to the internet because of accidental configuration changes I made ; in particular, [I removed](https://github.com/developmentseed/eoapi-cdk/compare/31a6845289f56f879ecb3b7a49ee38b9f6ee4c4c..01816fe874686444ad52230c86cb21bb34073325) `allowPublicSubnet: true` from the lambda creation parameters, as AWS explicitly says the lambda does not have access to the internet with this parameter. 

Doesn't seem to be solving the problem, though.

integration_tests/cdk/requirements.txt
```diff
@@ -0,0 +1,15 @@
+aws-cdk-lib>=2.99.1
+aws_cdk.aws_cognito_identitypool_alpha>=2.99.0a0
+aws-cdk.aws-apigatewayv2-alpha>=2.99.0a0
+constructs>=10.0.0,<11.0.0
```
vincentsarago, line 4: @emileten some requirements are not really needed because they are defined in eoapi-cdk 

https://github.com/developmentseed/eoapi-cdk/blob/1e39ddb54597754d4d621313e7ce34eebe9a8a6f/package.json#L53-L56

integration_tests/cdk/requirements.txt
```diff
@@ -0,0 +1,15 @@
...
+pydantic==2.0.2
+pydantic-settings==2.0.1
+black==22.3.0
+boto3==1.24.15
+boto3-stubs[cognito-idp,cognito-identity]
+flake8==4.0.1
+click==8.1.3
```
vincentsarago, line 11: click is not used

integration_tests/cdk/requirements.txt
```diff
@@ -0,0 +1,15 @@
+aws-cdk-lib>=2.99.1
+aws_cdk.aws_cognito_identitypool_alpha>=2.99.0a0
+aws-cdk.aws-apigatewayv2-alpha>=2.99.0a0
+constructs>=10.0.0,<11.0.0
+pydantic==2.0.2
+pydantic-settings==2.0.1
+black==22.3.0
+boto3==1.24.15
+boto3-stubs[cognito-idp,cognito-identity]
+flake8==4.0.1
```
vincentsarago, line 10: flake and black are dev dependencies

integration_tests/cdk/requirements.txt
emileten: Some of these, we don't use them explicitly in `integration_tests/cdk`, in which case I agree I should remove them. 

However shouldn't we still include those we _explicitly_ use, such as `constructs`: https://github.com/developmentseed/eoapi-cdk/blob/b4de0358eadd3cdabae3076ea4b8189b572d7bcd/integration_tests/cdk/eoapi_template/pgStacInfra.py#L6.

integration_tests/cdk/requirements.txt
emileten: thanks!

integration_tests/cdk/requirements.txt
emileten: just removed them ; we have this 'tox' workflow for linting/formatting.

lib/database/index.ts
```diff
@@ -60,7 +60,7 @@ export class PgStacDatabase extends Construct {
     });
 
     const handler = new aws_lambda.Function(this, "lambda", {
-      // defaults for configurable properties
```
emileten, line 63: here, and the same applies to all the changes to `aws_lambda.Function` calls, the only thing we're doing is making _all_ parameters passed to `aws_lambda.Function` configurable by the user. The way we're doing this is making sure we pass `props.[insert name]LambdaFunctionOptions` in the last position. Before, some network related options (notably vpc, allowPublicSubnet) were not configurable because they were hard coded at the end. 

Although this exposes users to errors if they pass conflictual parameter values (for example, they place a lambda in a vpc and in a public subnet without the allowPublicSubnet flag), it maximizes configurability.

PR #70, opened 2023-09-06
feat: update python version to 3.11

### LINKED ISSUE #68: Update runtimes to python 3.11
☝️

vincentsarago:
closes #68 

To Be discussed: should this be considered as a breaking change? @alukach @sharkinsspatial @emileten

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,18 +1,18 @@
 {
-  "name": "cdk-pgstac",
+  "name": "eoapi-cdk",
   "version": "5.4.0",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
-      "name": "cdk-pgstac",
+      "name": "eoapi-cdk",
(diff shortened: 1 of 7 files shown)


package-lock.json
```diff
@@ -1,18 +1,18 @@
 {
-  "name": "cdk-pgstac",
```
emileten, line 2: this still had `cdk-pgstac` instead of `eoapi-cdk` in it ? 😵‍💫

vincentsarago: oh it's interesting, I pushed my own package-lock.json while it seems that when doing a release only the version number was changed in the package-lock.json, not the whole file was update 🤷‍♂️

emileten:
FYI I'll be addressing this as part of https://github.com/developmentseed/eoapi-cdk/pull/74 -- I'll close here once the latter is merged.

vincentsarago:
done in #74

PR #73, opened 2023-09-18
fix: synchronize pgstac versions -> 0.7.9 across constructs

ividito:
Affects `stac-api`, `ingestor`, `bootstrapper`. These constructs have gotten out of sync with the `pgstac` versions used in other constructs. As far as I can tell, this only impacts the ingestor in a somewhat breaking way - prior to `pgstac` 0.7, `pypgstac` would fail if the major version of `pgstac` is mismatched with the loader. However, most constructs are already using updated dependencies. This PR should sync all constructs with `pgstac` 0.7.9.


emileten:
@ividito do you think this is ready for review ? Feel free to add reviewers (including me if you'd like!) if so

vincentsarago (approved):

PR #74, opened 2023-10-05
feat: custom runtimes, optional VPC, python 3.11

emileten:
A large PR with several breaking changes.

- ⚠️ breaking change. bootstrapper runtime made configurable. This necessitated to 'merge' the `database` and `bootstrapper` into a single construct to avoid JSII issues. With this, all apps have their runtime configurable. 

- ⚠️ breaking change. switched to `Function` and `fromDockerBuild` for lambda functions, because the state of `PythonFunction` makes it very hard to add flexible configurability. Now, users can override any of the lambda `Function` parameters they want ( code, but also memory, time out...).

- made VPC optional for lambdas
 
- harmonized pgstac to 0.7.10 in default runtimes. 

- harmonized python to 3.11 in default runtimes.


lib/bootstrapper/runtime/Dockerfile
```diff
@@ -1,11 +1,8 @@
 FROM lambci/lambda:build-python3.8
 
-ARG PGSTAC_VERSION
-RUN echo "Using PGSTAC Version ${PGSTAC_VERSION}"
-
 WORKDIR /tmp
 
-RUN pip install httpx psycopg[binary,pool] pypgstac==${PGSTAC_VERSION} -t /asset
+RUN pip install httpx psycopg[binary,pool] pypgstac==0.6.13 -t /asset
```
vincentsarago, line 5: Why is this hardcoded now?

vincentsarago: oh I see, now the user cannot `just` specify a pgstac version but has to provide a full docker build

vincentsarago (approved):
This looks good to me. 

We should make sure to document this part to explain how basic dockerfile should look

```dockerfile
# IMPORTANT: Use the same version of python for the asset building and lambda deployment
FROM public.ecr.aws/lambda/python:3.10

WORKDIR /tmp
RUN python -m pip install pip -U

# Install application dependency 
COPY runtime/requirements.txt requirements.txt
RUN python -m pip install -r requirements.txt "mangum>=0.14,<0.15" -t /asset

# Copy handler and custom code
COPY runtime/src/*.py /asset/
```

sharkinsspatial:
@emileten We should also consider adding [dockerignore file](https://codefresh.io/blog/not-ignore-dockerignore-2/) with `cdk.out` included at a minimum to reduce overall context size now that you are using Docker builds for all Lambdas.

lib/ingestor-api/index.ts
```diff
@@ -153,26 +158,28 @@ export class StacIngestor extends Construct {
...
+      ...props.lambdaFunctionOptions ?? {
+        runtime: lambda.Runtime.PYTHON_3_9,
+        handler: "handler.handler",
+        memorySize: 2048,
+        logRetention: aws_logs.RetentionDays.ONE_WEEK,
+        timeout: Duration.seconds(180)
+      },
+      code: props.lambdaAssetCode ?? lambda.Code.fromDockerBuild(__dirname, {
+        file: "runtime/Dockerfile",
+        buildArgs: { PYTHON_VERSION: '3.9' },
```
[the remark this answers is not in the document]
sharkinsspatial: It doesn't appear that any of your updated `Dockerfile`'s are actually leveraging `buildArgs`.

lib/bootstrapper/index.ts
```diff
@@ -28,17 +27,18 @@ export class BootstrapPgStac extends Construct {
   constructor(scope: Construct, id: string, props: BootstrapPgStacProps) {
     super(scope, id);
 
-    const { pgstacVersion = DEFAULT_PGSTAC_VERSION } = props;
     const handler = new aws_lambda.Function(this, "lambda", {
-      handler: "handler.handler",
-      runtime: aws_lambda.Runtime.PYTHON_3_8,
-      code: aws_lambda.Code.fromDockerBuild(__dirname, {
+      ...props.lambdaFunctionOptions ?? {
```
sharkinsspatial, line 31: Maybe someone with more TS experience than me can weigh in here (@alukach ?).  Is completely selecting either a fully specified `CustomLambdaFunctionOptions` interface or instead the manually specified "default" parameters the best approach for doing this?  

I would have guessed that using some combination of an interface with defaults and destructuring was more elegant but it has been a long time since I wrote any TS :]

lib/ingestor-api/index.ts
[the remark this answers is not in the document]
vincentsarago: @sharkinsspatial I think this is because now the user has to pass `aws_cdk.aws_lambda.Code` https://docs.aws.amazon.com/cdk/api/v2/python/aws_cdk.aws_lambda/Code.html so we can't really just pass build-arg as before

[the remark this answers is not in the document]
sharkinsspatial: @vincentsarago I might be misreading the TypeScript here but I think that construct users can optionally specify their own [lambda.assetCode](https://github.com/developmentseed/eoapi-cdk/blob/eb20644460d8f3f59339eb48bb3239cabea83860/lib/stac-api/index.ts#L31) or the construct will use `lambda.Code.fromDockerBuild` with the local runtime and `Dockerfile`.  Currently, the base image in the `Dockerfile` is not using the `buildArgs` that are passed.  They should be so that the base image can be specified dynamically to use the Python version set in the `runtime` parameter.

lib/bootstrapper/index.ts
```diff
@@ -28,17 +27,18 @@ export class BootstrapPgStac extends Construct {
...
+        runtime: aws_lambda.Runtime.PYTHON_3_8,
+        handler: "handler.handler",
+        memorySize: 128,
+        logRetention: aws_logs.RetentionDays.ONE_WEEK,
+        timeout: Duration.minutes(2)
+      },
+      code: props.lambdaAssetCode ?? aws_lambda.Code.fromDockerBuild(__dirname, {
```
sharkinsspatial, line 38: @emileten I'm wondering if we might want to support both `AssetCode` and `DockerImageCode` for the code prop to handle cases where user asset package builds might exceed 250MB.

lib/bootstrapper/index.ts
alukach: I think I would agree that this is possibly less ideal for two reasons:

1. It only allows users to specify _all_ options or partial.
2. We're using our own custom interface (`CustomLambdaFunctionOptions`) which isn't based on the actual [`aws_lambda.FunctionProps`](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_lambda.FunctionProps.html) that CDK uses. It's a subset of all the options a user may want to set and we will have to keep in sync with the lambda props as they change in future CDK versions.

I think a way around both of these would be for us to instead allow a user to submit none, some, or all the types specified in `aws_lambda.FunctionProps` and then apply defaults for properties that are not specified by the user.  I think we can achieve this via JS' destructuring (and the neat fact that you can destructure `undefined` without error) and TS' [`Partial` & `Omit` utility types](https://www.typescriptlang.org/docs/handbook/utility-types.html):

```ts

export class BootstrapPgStac extends Construct {
  // ...

  constructor(scope: Construct, id: string, props: BootstrapPgStacProps) {
    // ...

    const handler = new aws_lambda.Function(this, "lambda", {
      // Defaults for configurable properties
      runtime: aws_lambda.Runtime.PYTHON_3_8,
      handler: "handler.handler",
      memorySize: 128,
      timeout: Duration.minutes(2),

      // Configurable properties will overwrite defaults
      ...props.lambdaFunctionOptions,
      
      // Non-configurable
      code: props.lambdaAssetCode ?? aws_lambda.Code.fromDockerBuild(__dirname, {
        file: "runtime/Dockerfile",
      }),
      vpc: hasVpc(props.database) ? props.database.vpc : props.vpc,
    });
    
    // ...
  }
}


export interface BootstrapPgStacProps {
  //...

  // Allow user to specify none, some, or all props in AWS Lambda Function Props,
  // except for `code`.
  readonly lambdaFunctionOptions?: Partial<Omit<aws_lambda.FunctionProps, "code" | "vpc">>;

  //...
}

```

alukach: Thinking more about this, we can probably do away with the `lambdaAssetCode` and instead not omit the `code` property from `aws_lambda.FunctionProps`.

lib/ingestor-api/index.ts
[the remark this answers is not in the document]
vincentsarago: I agree that user should still be able to pass custom buildArgs even if we switch to full AssetCode

lib/bootstrapper/index.ts
sharkinsspatial: @emileten Is this something you want to try and tackle solo or should we hop in and try to tag team some revisions.

lib/ingestor-api/index.ts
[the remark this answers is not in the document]
emileten: @sharkinsspatial

> It doesn't appear that any of your updated Dockerfile's are actually leveraging buildArgs.

This is correct, and I fixed this mistake last friday but hadn't pushed . That said, your idea here : 

> dynamically specifying the base Lambda image's Python version

That would be great, but only if it's easy to use. There seems to be just that [exposed by the CDK](https://github.com/aws/aws-cdk/issues/26372). Could try that.

lib/bootstrapper/index.ts
emileten: Just looked into this in detail (was off on monday), I totally agree that : 

> It only allows users to specify all options or none.
>  We're using our own custom interface

Is not good, I just had no way around this, but @alukach guidance sounds great, trying to implement this.

lib/bootstrapper/index.ts
emileten: @sharkinsspatial `code` in `aws_lambda.Function` must be of [type `aws_lambda.Code`](https://github.com/aws/aws-cdk/blob/5fe46d6f6ce325caa9a62ee5bb015821068e77ee/packages/aws-cdk-lib/aws-lambda/lib/code.ts#L12), and `DockerImageCode` doesn't seem to be implementing this. 

You use `DockerImageCode` with `aws_lambda.DockerImageFunction`instead. One way to do this would be, given the change suggested by Anthony : 
- detect if in `lambdaFunctionOptions` the user passed a value to `code`. 
- If there is, check its type. If it's `DockerImageCode`, then we use `DockerImageFunction`, otherwise and also if it's absent, we fall back to `Function`. 
  
So that adds an (albeit small) layer of complexity that I'd maybe rather tackle later in the absence of an immediate need.

lib/bootstrapper/index.ts
emileten: Unfortunately, I think we won't be able to do that because those are not supported by JSII @alukach, see https://github.com/aws/jsii/issues/2901. 

My suggested workaround is to say that the user can pass anything from the aws_lambda.FunctionProps type but raise an error if it's vpc/subnet/role etc..., and say it in the docs. It's imperfect because e.g. the user will not be warned by the type checker, but better than copy/pasting properties which is what i was doing initially.

emileten: side note, looks like the repo this issue originated from (maintained by aws) serves the same purpose as ours -- defining shareable constructs --  but for typical multi-aws-service deployments. Could be a source of inspiration !

https://github.com/awslabs/aws-solutions-constructs

emileten: see https://github.com/developmentseed/eoapi-cdk/pull/74/commits/dbcada8cf65e3264a71144c660e9a688db9f1e59

lib/ingestor-api/index.ts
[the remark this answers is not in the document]
emileten: Unfortunately, like mentioned by an aws-cdk maintainer in [this issue](https://github.com/aws/aws-cdk/issues/26372), this feature seems to be supported for `lambda.DockerImageFunction` only.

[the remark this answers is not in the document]
emileten: > I agree that user should still be able to pass custom buildArgs even if we switch to full AssetCode

that's going to be the case now, see https://github.com/developmentseed/eoapi-cdk/commit/dbcada8cf65e3264a71144c660e9a688db9f1e59

emileten:
I addressed comments ; I'll see if I can fit that into [eoAPI](https://github.com/developmentseed/eoAPI/blob/main/infrastructure/aws/cdk/app.py) and after confirming it works I'll merge.

emileten:
@vincentsarago FYI in the meanwhile (this took me forever 🙄) I did these changes (I updated the description): 

- -> python 3.11 (so I'll close your other PR)
- made VPC optional for lambdas (this allows for a cheap, low-security deployment without a nat gateway)
- to make the bootstrapper runtime customizable I had to merge the database and the boostrapper construct together. This is a breaking change but unfortunately necessary, the things were designed with the two constructs interdependent but separated was causing JSII issues. 

[This is tested in deployment](https://github.com/developmentseed/eoAPI/pull/144) and good to go, about to merge!

vincentsarago (approved):

PR #80, opened 2023-10-26
fix misleading error-like message

hrodmn:
This bug doesn't break deployment but it prints a confusing error-like message!

```sh
send(..) failed executing httpx.put(..): 'Response' object has no attribute 'reason'
```

### DIFF
--- a/lib/bootstrapper/runtime/handler.py
+++ b/lib/bootstrapper/runtime/handler.py
@@ -59,7 +59,7 @@ def send(
 
     try:
         response = httpx.put(responseUrl, data=json_responseBody, headers=headers)
-        print("Status code: " + response.reason)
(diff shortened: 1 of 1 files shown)


vincentsarago:
👍 

FYI: in eoapi I've replaced `reason` with `status_code` 

https://github.com/developmentseed/eoAPI/commit/3efc167c06d8b84b81e9294094715b4051d2559f#diff-c13e2797ef93f382fbaa2bce6d35da917665c4f48d90598c7a86c8e27b97226aR64

emileten:
Thank you for raising this @hrodmn !

Looks like `response_code`is not an available attribute. [I opened a PR](https://github.com/developmentseed/eoapi-cdk/pull/81) that uses `status_code` instead.

PR #86, opened 2023-11-23
add params for package versions

### LINKED ISSUE #85: add config settings for package versions (e.g. titiler-pgstac==0.5.1)
Right now the specific versions of the critical python packages are hard-coded in the runtime files

e.g.
https://github.com/developmentseed/eoapi-cdk/blob/1e39ddb54597754d4d621313e7ce34eebe9a8a6f/lib/titiler-pgstac-api/runtime/requirements.txt#L1

I have been following along over in https://github.com/developmentseed/eoAPI/pull/144 and see pattern you follow for overriding the runtime. I suppose I could use this method to achieve my goal but it seems heavy-handed for my goal of staying up-to-date on dependency upgrades!

https://github.com/developmentseed/eoAPI/pull/144/files#diff-5e565e714fe20093460499e9958e38735b99628ac37b603c30e2c8af61aee726R169-R184

One of the things that is very convenient about eoapi-cdk is that it makes it possible to just import the cdk modules without maintaining separate runtime scripts, so adding some configurability for the deployment of the packages would be nice!

Would it make sense to be able to set the versions of packages like titiler-pgstac and tipg as settings in the config? Maybe it would be tricky since some of the handler code might be dependent on version-specific details, but it seems like a way to decouple versions of eoapi-cdk from versions of the family of dependencies.

hrodmn:
## What I am changing

- make it possible to configure the versions of these packages without custom runtimes
  - tipg
  - titiler-pgstac
  - stac-fastapi
  - pgstac

## How I did it

- added a constant to the top of each relevant typescript file with the default version as specified in v6.1.0
- added new arguments to the relevant constructs
  - `TitilerPgstacApiLambda`: `titiler_pgstac_version`
  - `PgStacApiLambda`: `stac_fastapi_version`
  - `PgStacDatabase`: `pgstac_version`
  - `TiPgApiLambda`: `tipg_version`
- used docker build args to specify the versions at build time

## How you can test it

- run an eoapi-template deployment with no versions specified - this should yield an identical deployment to v6.1.0
- run an eoapi-template deployment with user-specified versions of titiler-pgstac, tipg, pgstac, stac-fastapi
  ```python
  # STAC DB
  pgstac_db = PgStacDatabase(
      self,
      "pgstac-db",
      pgstac_version="0.8.2",
      ...
  )

  # titiler-pgstac
  TitilerPgstacApiLambda(
      self,
      "titiler-pgstac-api",
      titiler_pgstac_version="0.7.10",
      ...
  )

  # stac-fastapi
  stac_api_lambda = PgStacApiLambda(
      self,
      "pgstac-api",
      stac_fastapi_version="2.4.9",
      ...
  )

  # tipg
  TiPgApiLambda(
      self,
      "tipg-api",
      tipg_version="0.5.1",
      ...
  )
  ```

## Related Issues

- resolves #85

### DIFF
--- a/lib/database/index.ts
+++ b/lib/database/index.ts
@@ -68,7 +68,7 @@ export class PgStacDatabase extends Construct {
       timeout: Duration.minutes(2),
       code: aws_lambda.Code.fromDockerBuild(__dirname, {
         file: "bootstrapper_runtime/Dockerfile",
-        buildArgs: {PGSTAC_VERSION: DEFAULT_PGSTAC_VERSION, PYTHON_VERSION: "3.11"}
+        buildArgs: {PGSTAC_VERSION: props.pgstacVersion || DEFAULT_PGSTAC_VERSION, PYTHON_VERSION: "3.11"}
       }),
       // overwrites defaults with user-provided configurable properties
       ...props.bootstrapperLambdaFunctionOptions,
@@ -120,7 +120,7 @@ export class PgStacDatabase extends Construct {
 
     // if props.lambdaFunctionOptions doesn't have 'code' defined, update pgstac_version (needed for default runtime)
     if (!props.bootstrapperLambdaFunctionOptions?.code) {
-      customResourceProperties["pgstac_version"] = DEFAULT_PGSTAC_VERSION;
+      customResourceProperties["pgstac_version"] = props.pgstacVersion || DEFAULT_PGSTAC_VERSION;
     }
     // this.connections = props.database.connections;
     new CustomResource(this, "bootstrapper", {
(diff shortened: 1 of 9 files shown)


hrodmn:
I tried launching a stack with `pgstac==0.8.1` and got an error in the bootstrap process:
```sh
...
Running migrations...
Unable to bootstrap database with exception=must be member of role "pgstac_ingest"
```

python traceback:
```python
  File "/var/task/handler.py", line 232, in handler
    raise e
  File "/var/task/handler.py", line 205, in handler
    Migrate(pgdb).run_migration(params["pgstac_version"])
  File "/var/task/pypgstac/migrate.py", line 149, in run_migration
    cur.execute(migration_sql)
  File "/var/task/psycopg/cursor.py", line 737, in execute
    raise ex.with_traceback(None)
```

I think this is a case where the handler code might need to be updated to accommodate changes in one of the dependencies (`pgstac`). Would it make sense to restrict to particular major/minor versions to prevent errors like this? In this case, `pgstac==0.7.10` works fine but `pgstac==0.8.2` does not.

We could use the `semver` package to check that the supplied version of `pgstac` is `<0.8`. I am not exactly sure how we would set the limits for other packages but maybe something to consider for `pgstac` right now.

hrodmn:
I'm closing this after more discussion in #85 that suggests we don't want to go down this road.

PR #91, opened 2024-01-17
build: add dependabot for runtimes

vincentsarago:
This PR aims to add Dependabot to automatically open PR to update runtimes.

I think we would need to customize commit message so it does release. 

I'm not 💯 convince about this workflow though because we have 3 diff Dependabot which could in theory open 3 different PR so each merge could create 3 diff releases 😬


chris-bateman:
@vincentsarago Assuming I understand how you do releases correctly :)

This would theoretically solve the issue, but it's still an open task.
https://github.com/dependabot/dependabot-core/issues/8126

In the meantime, maybe make a group for each ecosystem to ensure you only get 1 PR per ecosystem and then set the [prefix](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#commit-message) so that you only do a fix/patch bump?  It could still be three versions, but at least what changed in each one is clear.

vincentsarago:
> In the meantime, maybe make a group for each ecosystem to ensure you only get 1 PR per ecosystem and then set the [prefix](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#commit-message) so that you only do a fix/patch bump? It could still be three versions, but at least what changed in each one is clear.

Yes I should do this 🙏

hrodmn:
superseded by https://github.com/developmentseed/eoapi-cdk/pull/260

PR #96, opened 2024-02-22
chore: simplify deployment

emileten:
- [x] The manual deployment workflow ran successfully

- have all the cdk code in one `app.py` file 
- add the vector app
- add a letter representation of `datetime.datetime.utcnow()` to the test stack ids in _case_ (who knows 🤷‍♂️) several people try to test their PR at the same time. This will fail though if they trigger their workflow at the exact same second, but that's much less likely to happen ! I could use the branch name instead, but then I need to change the workflow again... _edit: eventually decided to use the branch name_ _edit: second edit, branch name has slashes sometimes, which doesn't work with cloudformation, so I decided to use the short commit sha eventually.


emileten:
workflow running https://github.com/developmentseed/eoapi-cdk/actions/runs/8004317622/job/21861453171

emileten:
Fun times ! The workflow 'succeeded', but the stack wasn't actually destroyed. This is causing me challenges : 

```
      - name: Tear down any infrastructure
        if: always()
        run: |
          cd integration_tests/cdk
          # run this only if we find a 'cdk.out' directory, which means there might be things to tear down
          if [ -d "cdk.out" ]; then
            cd -
            source .deployment_venv/bin/activate
            cd integration_tests/cdk
            # see https://github.com/aws/aws-cdk/issues/24946
            # this didn't work : rm -f cdk.out/synth.lock
            # so we just duplicate the cdk output to cdk-destroy.out
            npx cdk destroy --output cdk-destroy.out --ci --all --force
          fi
```

I want the workflow to be resilient to a sudden cancel in the middle of the deployment ; i.e, I want the _destruction_ to happen properly in these cases. The problem is that in these cases, `cdk.out/synth.lock` is not released by the deployment step, which probably runs in a docker container. [This is a known issue.](https://github.com/aws/aws-cdk/issues/24946).

People have been saying that removing the file manually `rm cdk.out/synth.lock` solves this, but it did not for me. So, I instead decided to have the synthesized output of `cdk destroy` be saved in a _different_ directory, like above : 

```
npx cdk destroy --output cdk-destroy.out --ci --all --force
```

This is probably why the stack doesn't _actually_ get destroyed...

vincentsarago (approved):
🚀

emileten:
successful workflow : https://github.com/developmentseed/eoapi-cdk/actions/runs/8010980629 merging

PR #100, opened 2024-02-26
feat: update runtimes and add pgstac customization options

### LINKED ISSUE #88: Update core packages in default runtimes
Some of the core packages are quite out of date in our default deployment : 

- latest tipg is 0.5.1 but ours is 0.3.0 https://github.com/developmentseed/eoapi-cdk/blob/1e39ddb54597754d4d621313e7ce34eebe9a8a6f/lib/tipg-api/runtime/Dockerfile#L7
- latest titiler-pgstac is 0.8.1, ours is 0.5.1
- latest pgstac is 0.8.2, ours is 0.7.10.

We should update these.

vincentsarago:
This PR closes #88 and add customizations option for PgSTAC bootstrap to:
- enable/disable the *context* extension 
- enable/disable the *mosaic* table index

## Merge request description


lib/database/bootstrapper_runtime/handler.py
```diff
@@ -140,6 +145,33 @@ def register_extensions(cursor) -> None:
...
+        ON CONFLICT ON CONSTRAINT pgstac_settings_pkey DO UPDATE SET value = excluded.value;"""
+        cursor.execute(sql.SQL(pgstac_settings))
+
+    if params.get("mosaic_index", False):
+        # Create index of searches with `mosaic`` type
+        cursor.execute(
+            sql.SQL(
+                "CREATE INDEX IF NOT EXISTS searches_mosaic ON searches ((true)) WHERE metadata->>'type'='mosaic';"
+            )
+        )
```
vincentsarago, line 172: this is taken from eoapi repo

lib/database/bootstrapper_runtime/handler.py
```diff
@@ -203,12 +235,12 @@ def handler(event, context):
...
-        print(f"Current {pgdb.version=}")
+        with PgstacDB(dsn=stac_db_admin_dsn, debug=True) as pgdb:
+            print(f"Current {pgdb.version=}")
 
-        # As admin, run migrations
-        print("Running migrations...")
-        Migrate(pgdb).run_migration(params["pgstac_version"])
+            # As admin, run migrations
+            print("Running migrations...")
+            Migrate(pgdb).run_migration(params["pgstac_version"])
```
vincentsarago, line 243: cleaner way to make sure the connection gets closed

lib/database/bootstrapper_runtime/handler.py
```diff
@@ -220,17 +252,27 @@ def handler(event, context):
...
+            "postgresql://{user}:{password}@{host}:{port}/{dbname}".format(
+                dbname=user_params.get("dbname", "postgres"),
+                user=user_params["username"],
+                password=user_params["password"],
+                host=connection_params["host"],
+                port=connection_params["port"],
             )
+        )
+        with PgstacDB(dsn=stac_db_user_dsn, debug=True) as pgdb:
+            print(f"User can access pgstac: {pgdb.version}")
```
vincentsarago, line 275: we had this in eoapi repo which I think makes sense to test

emileten:
deployment workflow https://github.com/developmentseed/eoapi-cdk/actions/runs/8045501396

lib/tipg-api/runtime/requirements.txt
```diff
@@ -0,0 +1 @@
+tipg=0.6.3
```
emileten, line 1: there is a bug here I think, we should pbbly change to 👇 
```suggestion
tipg==0.6.3
```

vincentsarago: good catch

lib/database/bootstrapper_runtime/handler.py
```diff
@@ -66,8 +69,10 @@ def send(
             timeout=30,
         )
         print("Status code: " + response.status_code)
+        logger.debug(f"OK - Status code: {response.status_code}")
     except Exception as e:
         print("send(..) failed executing httpx.put(..): " + str(e))
+        logger.debug(f"NOK - failed executing PUT requests:  {e}")
```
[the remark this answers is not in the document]
vincentsarago: well I was hoping to get this printed somewhere in cloudwatch 😅 we can remove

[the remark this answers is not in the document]
emileten: this does end up in cloud watch with just the `print`, in the lambda's logs!

emileten:
Questions for my understanding 

>  enable/disable the context extension
    enable/disable the mosaic table index

What were we doing before, was it disabled by default?

lib/database/bootstrapper_runtime/handler.py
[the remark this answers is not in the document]
emileten: (as far as I experienced -- was it impossible to find in your case?)

vincentsarago:
@emileten we were adding the mosaic index by default but not the context extension, maybe we can switch back to false 👍

lib/database/bootstrapper_runtime/handler.py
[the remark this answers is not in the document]
vincentsarago: yeah I've never found any logs 😅

emileten:
> Maybe we can switch back to false

I'll defer to you on this decision, I don't know much about these options. Doesn't sound like it hurts to have the context on by default ?

lib/tipg-api/runtime/requirements.txt
emileten: the deployment action caught it. It's useful at least 😄

lib/database/index.ts
```diff
@@ -14,7 +14,12 @@ import { Construct } from "constructs";
 import { CustomLambdaFunctionProps } from "../utils";
 
 const instanceSizes: Record<string, number> = require("./instance-memory.json");
-const DEFAULT_PGSTAC_VERSION = "0.7.10";
+const DEFAULT_PGSTAC_VERSION = "0.8.4";
```
emileten, line 17: If we do that, we need to change the `pypgstac` pin in the ingestor https://github.com/developmentseed/eoapi-cdk/blob/a5e398e596fcb80398db10cfdb87aad9e20673bf/lib/ingestor-api/runtime/requirements.txt#L8

lib/database/bootstrapper_runtime/handler.py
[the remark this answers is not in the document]
emileten: Hm that's not good.. [We do have a CloudWatch endpoint at least in the tests](https://github.com/developmentseed/eoapi-cdk/blob/a5e398e596fcb80398db10cfdb87aad9e20673bf/integration_tests/cdk/app.py#L43), so it shouldn't be caused by a connectivity issue ?

For me it takes a couple minutes but usually I find them. E.g. when `httpx.put` timed out the other day. 

it doesn't hurt to have both, I am fine leaving them.

emileten:
A live example of the bootstrapper logs

The workflow failed again, there's something we need to fix in the bootstrapper changes https://github.com/developmentseed/eoapi-cdk/actions/runs/8045602231/job/21971192880#step:9:2478

The lambda logs https://us-east-1.console.aws.amazon.com/cloudwatch/home?region=us-east-1#logsV2:log-groups/log-group/$252Faws$252Flambda$252Fpgstac60fe14-pgstacdblambda91737152-ELDP6T9rb8jI/log-events/2024$252F02$252F26$252F$255B$2524LATEST$255Df9a0fff7fa2042bb84e34c9220cb7071

```
Unable to bootstrap database with exception=must be member of role "pgstac_ingest" 
```

Also `httpx` failed responding with the error

```
send(..) failed executing httpx.put(..): can only concatenate str (not "int") to str 
```

emileten:
@vincentsarago just checking, were you expecting the last commit to fix the bug ? or is there more to do

emileten:
ugh sorry, just started one. Will show you later

vincentsarago:
@emileten let's fall back to pgstac 0.7.10 to validate the changes and then we can debug why it doesn't work for pgstac 0.8.4

emileten:
@vincentsarago like this https://github.com/developmentseed/eoapi-cdk/pull/100/commits/0d675e21d782cdefa7697d5310d6e6aece21a866?

vincentsarago:
@emileten just pgstac, we can keep titiler-pgstac and tipg to the latest version

emileten:
Ok seems to work https://github.com/developmentseed/eoapi-cdk/actions/runs/8204255860/job/22438560336#step:10:798. 

Now we need to 

>  debug why it doesn't work for pgstac 0.8.4

😄

emileten:
there seems to be a 0.8.3 change that sounds like it could have triggered this https://github.com/stac-utils/pgstac/blob/e0b7de4c9c8130c55de9fe6eb0a87324cd705135/CHANGELOG.md?plain=1#L37 we can try with the previous version

vincentsarago:
👁️ https://github.com/stac-utils/pgstac/issues/239

vincentsarago:
@emileten if I understand well 

looking at https://github.com/developmentseed/eoapi-cdk/pull/100#issuecomment-1983778135 it seems the `migration` step fails with 

```
``` 
in https://github.com/developmentseed/eoapi-cdk/blob/44252534b55ce3266add771ec1f4de7c6848e53e/lib/database/bootstrapper_runtime/handler.py#L243

when we use 

https://github.com/developmentseed/eoapi-cdk/blob/44252534b55ce3266add771ec1f4de7c6848e53e/lib/database/bootstrapper_runtime/handler.py#L231-L232

username/password (must be `postgres` by default) 

After the migration we then grant the username/password (`pgstac_user` by default) 

https://github.com/developmentseed/eoapi-cdk/blob/44252534b55ce3266add771ec1f4de7c6848e53e/lib/database/bootstrapper_runtime/handler.py#L133-L135

so to me it seems that we should `GRANT pgstac_ingest TO postgres;` but to me it seems weird because `pgstac_ingest` role doesn't exist before the migration starts... 

cc @bitner

bitner:
@emileten @vincentsarago 
The pypgstac migrate should be run as superuser (or a role with sufficient root-like privileges in the case of systems like RDS that do not allow true superusers -- this is typically the postgres role). `pypgstac migrate` on a fresh database will create the pgstac_read, pgstac_ingest, and pgstac_admin roles.

Once pgstac **has been installed**, you can assign these roles to other database login roles. Ideally, you would use least-access necessary roles for different tasks. Best practice would be to have three roles to interact with pgstac.

- An admin role that is used to manage migrations. Whatever role that is used to run pypgstac migrate will automatically be granted the pgstac_admin role.
- A loader role that is used with the transactions extension or when using `pypgstac load`. `GRANT pgstac_ingest to dataloader;`
- A user role that is used with the API for search or other read operations `GRANT pgstac_read to apiuser;`. Do note that this role is not truly read only as it will still be able to (indirectly via functions that run as the pgstac_admin role) update the searches and search_wheres table.

PR #102, opened 2024-03-12
(fix): fix database bootstrap

vincentsarago:

## Merge request description


lib/database/bootstrapper_runtime/handler.py
```diff
@@ -181,94 +181,95 @@ def handler(event, context):
...
-            user=connection_params["username"],
-            password=connection_params["password"],
-            host=connection_params["host"],
-            port=connection_params["port"],
+
+        # Admin (AWS RDS) user/password/dbname parameters
+        admin_params = get_secret(params["conn_secret_arn"])
+
+        # Custom eoAPI user/password/dbname parameters
+        eoapi_params = get_secret(params["new_user_secret_arn"])
```
vincentsarago, line 189: Tried to make the variable names a bit more clear!

lib/database/bootstrapper_runtime/handler.py
```diff
@@ -56,20 +52,18 @@ def send(
...
         response = httpx.put(
-            responseUrl,
+            event["ResponseURL"],
             data=json_responseBody,
-            headers=headers,
+            headers={"content-type": "", "content-length": str(len(json_responseBody))},
             timeout=30,
         )
-        print("Status code: " + response.status_code)
+        print("Status code: ", response.status_code)
```
vincentsarago, line 64: there was a bug here

lib/database/bootstrapper_runtime/handler.py
```diff
@@ -235,35 +229,25 @@ def handler(event, context):
...
+            print("Customize PgSTAC database...")
             # Update permissions to eoAPI user to assume pgstac_* roles
             with conn.cursor() as cur:
-                print(f"Update *{eoapi_params['username']}* permissions...")
+                print(f"Update '{eoapi_params['username']}' permissions...")
                 update_user_permissions(
                     cursor=cur,
                     db_name=eoapi_params["dbname"],
                     username=eoapi_params["username"],
                 )
```
vincentsarago, line 245: we need to have the search path set to pgstac,public in order to update the user permission or granting `pgstac_*`  to user won't work 😅

bitner: One trick that might be useful is to just set the search_path on the new database you are creating.
`ALTER DATABASE mydb SET SEARCH_PATH TO pgstac, public;`

vincentsarago:
ok, So deployment still fails with the same error message 
```
  File "/var/task/handler.py", line 266, in handler
    raise e
  File "/var/task/handler.py", line 230, in handler
    Migrate(pgdb).run_migration(params["pgstac_version"])
  File "/var/task/pypgstac/migrate.py", line 149, in run_migration
    cur.execute(migration_sql)
  File "/var/task/psycopg/cursor.py", line 732, in execute
    raise ex.with_traceback(None)
```

When running the `handler` locally I don't get any error! there must be something special about RDS

PR #105, opened 2024-05-14
feat: overwrite host headers when custom domain name

### LINKED ISSUE #75: Add custom parameter mapping when using custom API domain
When using custom API domain in API Gateway the FastAPI application (starlette) will still receive the api gateway hostname and thus use it internally (e.g if you have an endpoint that construct other endpoint url, it will return the API gateway url instead of you custom domain name)

e.g titiler 
```python
import httpx

resp = httpx.get("https://stac.eoapi.dev")
print(resp.json()["links"][2]

>>      {
      "rel": "data",
      "type": "application/json",
      "href": "https://stac.eoapi.dev/collections"
    }
```

☝️ this is giving the correct result (stac.eoapi.dev hostname) because I've manually added custom parameter mapping in the API Gateway console:

Screenshot 2023-10-05 at 1 00 59 PM

I believe we can automatically create this in 👇 

https://github.com/developmentseed/eoapi-cdk/blob/a1081a22e034d50f89d248d36844d598d45d5c69/lib/stac-api/index.ts#L60-L65

with: 
https://constructs.dev/packages/@aws-cdk/aws-apigatewayv2/v/1.204.0/api/ParameterMapping?lang=typescript

```js
    const stacApi = new HttpApi(this, `${Stack.of(this).stackName}-stac-api`, {
      defaultDomainMapping: props.stacApiDomainName ? { 
        domainName: props.stacApiDomainName
      } : undefined,
      defaultIntegration: new HttpLambdaIntegration("integration", this.stacApiLambdaFunction, parameterMapping = props.stacApiDomainName ? {new ParameterMapping().overwriteHeader('host', MappingValue.requestHeader(props.stacApiDomainName))} : undefined),
    });
```

vincentsarago:
closes #75

### DIFF
--- a/lib/stac-api/index.ts
+++ b/lib/stac-api/index.ts
@@ -8,7 +8,7 @@ import {
   Duration,
   aws_logs,
 } from "aws-cdk-lib";
-import { IDomainName, HttpApi } from "@aws-cdk/aws-apigatewayv2-alpha";
+import { IDomainName, HttpApi, ParameterMapping, MappingValue} from "@aws-cdk/aws-apigatewayv2-alpha";
 import { HttpLambdaIntegration } from "@aws-cdk/aws-apigatewayv2-integrations-alpha";
 import { Construct } from "constructs";
 import { CustomLambdaFunctionProps } from "../utils";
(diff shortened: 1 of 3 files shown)


alukach (approved):
Nice solution. 🎉

PR #112, opened 2024-12-09
fix: update units for effectiveCacheSize and tempBuffers

hrodmn:

## Merge request description
The `effectiveCacheSize` and `tempBuffers` parameters need to be described in number of 8kb blocks like `sharedBuffers`.
 https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/Appendix.PostgreSQL.CommonDBATasks.Parameters.html

Successful deployment workflow run: https://github.com/developmentseed/eoapi-cdk/actions/runs/12241535813

### DIFF
--- a/lib/database/index.ts
+++ b/lib/database/index.ts
@@ -159,11 +159,11 @@ export class PgStacDatabase extends Construct {
     return {
       maxConnections: `${maxConnections}`,
       sharedBuffers: `${sharedBuffers / 8}`, // Represented in 8kb blocks
-      effectiveCacheSize: `${effectiveCacheSize}`,
+      effectiveCacheSize: `${effectiveCacheSize / 8}`, // Represented in 8kb blocks
       workMem: `${workMem}`,
       maintenanceWorkMem: `${maintenanceWorkMem}`,
       maxLocksPerTransaction: "1024",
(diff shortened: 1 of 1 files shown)


alukach (approved):
This matches what the documentation describes (thanks for linking).

How did you come across this issue? Did this cause you problems?

Is it safe to say that existing `eoapi-cdk` deployments' `effective_cache_size` and `temp_buffers` are thus over-provisioned 8-fold?

hrodmn:
Thanks for the review @alukach!

> How did you come across this issue? Did this cause you problems?

I don't know if it is the source of my the issue, but I have been trying to improve performance on the MAAP eoAPI deployment by adding a `pgbouncer` instance between the titiler-pgstac Lambda and the RDS instance. While trying to determine some good `pgbouncer` settings for the `t3.micro` RDS instance, I was looking at the database configuration parameters and saw that `effective_cache_size` was 6GB for my 1GB instance!

> Is it safe to say that existing eoapi-cdk deployments' effective_cache_size and temp_buffers are thus over-provisioned 8-fold?

Yes I think any deployment that is not overriding the default has effective_cache_size that is 8x higher than we want.

PR #114, opened 2025-01-03
feat: add pgbouncer

hrodmn:
Add option to set up a pgbouncer server that can manage traffic to the actual database


hrodmn:
@vincentsarago I have it set up to work with Lambdas in or out of the VPC. Now the `pgbouncer` instance will advertise its public IP address if the database instance is in a public subnet and use its private ip address if the database is in a private subnet. Lambdas in the vpc can use the private ip address while Lambdas outside the vpc need to use the public ip address.

I added some more notes about networking setups over here: #116.

vincentsarago (approved):

PR #123, opened 2025-02-04
feat: configure pgstac version in custom properties

vincentsarago:
enable configuration of pypgstac version in custom ressource properties

### DIFF
--- a/lib/database/bootstrapper_runtime/Dockerfile
+++ b/lib/database/bootstrapper_runtime/Dockerfile
@@ -1,13 +1,11 @@
 ARG PYTHON_VERSION
 FROM --platform=linux/amd64 public.ecr.aws/lambda/python:${PYTHON_VERSION}
 
-ARG PGSTAC_VERSION
-RUN echo "PGSTAC_VERSION: ${PGSTAC_VERSION}"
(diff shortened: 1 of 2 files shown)


lib/database/bootstrapper_runtime/Dockerfile
```diff
@@ -1,13 +1,11 @@
...
 FROM --platform=linux/amd64 public.ecr.aws/lambda/python:${PYTHON_VERSION}
 
-ARG PGSTAC_VERSION
-RUN echo "PGSTAC_VERSION: ${PGSTAC_VERSION}"
 RUN echo "PYTHON_VERSION: ${PYTHON_VERSION}"
 
 WORKDIR /tmp
 
-RUN pip install httpx psycopg[binary,pool] pypgstac==${PGSTAC_VERSION} -t /asset
+RUN pip install httpx psycopg[binary,pool] pypgstac -t /asset
```
vincentsarago, line 8: we don't really need to set the pypgstac version here. In theory pypstac 0.9.2 should be able to run the migration step for 0.8

hrodmn: good idea - so as long as we install the latest version the bootstrapper should be able to do any migration.

hrodmn (approved):
:rocket:

PR #124, opened 2025-02-05
fix: ensure db bootstrapper runs on each deploy

### LINKED ISSUE #113: Updates to database bootstrapper Lambda do not re-trigger the function
I recently deployed a `PgstacDatabase` and discovered that even though the bootstrapper code had changed, the bootstrap script did not get re-triggered like I expected. I think this is because the `CustomResource` that triggers the function will only run if one of its properties changes.
https://github.com/developmentseed/eoapi-cdk/blob/9224d0a837f6f90aac0103d059c29a4bc7c7fa5a/lib/database/index.ts#L114-L123

Maybe we could add a property with a hash from the contents of the `bootstrapper_runtime` that would force the bootstrapper to run again.

hrodmn:
fix: ensure db bootstrapper runs on each deploy

fix: advertise secretBootstrapper so other resources can use it as a dependency

## Merge request description
This ensures that the bootstrapper `CustomResource` will run on every deploy, which will ensure that any changes to the bootstrap process get executed rather than just updating the Lambda and not invoking it. resolves #113.

I also added the `secretBootstrapper` attribute so other constructs can add it as a dependency (resolves #123)

### DIFF
--- a/integration_tests/cdk/app.py
+++ b/integration_tests/cdk/app.py
@@ -81,18 +81,16 @@ class pgStacInfraStack(Stack):
             instance_type=aws_ec2.InstanceType(app_config.db_instance_type),
             add_pgbouncer=True,
             removal_policy=RemovalPolicy.DESTROY,
+            pgstac_version="0.9.2",
         )
 
         assert pgstac_db.security_group
-
-        # make sure we can get the secret value!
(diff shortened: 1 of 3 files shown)


vincentsarago (approved):

vincentsarago:
we need to pin pydantic-ssm-settings ~=0.2

PR #128, opened 2025-03-03
feat: update versions for pgstac, stac-fastapi-pgstac, titiler-pgstac, and tipg

### LINKED ISSUE #115: Upgrade default versions of pgstac, titiler-pgstac, stac-fastapi-pgstac, tipg
We should upgrade the defaults to the latest versions of the main packages:
- `pgstac==0.8.5` -> `0.9.2`
- `stac-fastapi.pgstac==2.4.8` -> `3.0.1`
- `titiler.pgstac==1.2.2` -> `1.3.0`
- `tipg==0.6.3` -> `0.8.0`

hrodmn:
- [x] Posted the link to a successful manually triggered deployment workflow (successful including the resources destruction) https://github.com/developmentseed/eoapi-cdk/actions/runs/13770174796

I also manually created a deployment with `integration_tests/cdk/app.py` and verified that all of the services are working with these changes.

## Merge request description
Update versions for the core dependencies! Should this be a major version change since there are breaking changes in the APIs for stac-fastapi-pgstac (stac-fastapi>=0.5) and titiler-pgstac (from titiler>=0.19)?

Resolves #115 

I don't love using `os.environ.update(...)` to set the postgres settings in the handlers, but that is the only way to load e.g. `stac_fastapi.pgstac.app` since that module will load settings from the environment and fail if the `POSTGRES_*` variables are not set in the environment. We can load these settings in the new `config` modules, but to use the custom settings we have to completely re-define the app module in this repo - I would much rather use the pre-defined `app` variables here to make long-term maintenance easier.

@vincentsarago what do you think? We could refactor titiler-pgstac and stac-fastapi-pgstac a little bit so the application modules can be loaded even if those environment variables are present, then fail loudly on application startup if they are not configured.

One option would be to update the Lambda constructs with those `POSTGRES_*` environment variables set directly rather than setting `PGSTAC_SECRET_ARN`. I know we used to do that in eoapi-devseed but maybe it's not ideal from a security standpoint.


hrodmn:
I opted to remove all of the custom `app` and `config` modules in favor of the default applications that get shipped with `stac_fastapi.pgstac` and `titiler.pgstac`. This will make updating the default versions simpler in eoapi-cdk in the future and hopefully reduce confusion for users. You either get the default application or you can write a custom runtime.

lib/stac-api/runtime/src/handler.py
```diff
@@ -6,8 +6,40 @@
...
+
+from stac_fastapi.pgstac.app import app
+from stac_fastapi.pgstac.db import close_db_connection, connect_to_db
+
+
+@app.on_event("startup")
+async def startup_event():
+    """Connect to database on startup."""
+    print("Setting up DB connection...")
+    await connect_to_db(app)
```
vincentsarago, line 32: Sadly settings and `connect_to_db` are defined a bit differently in stac-fastapi-pgstac https://github.com/stac-utils/stac-fastapi-pgstac/blob/b1b393c6d283da0dd4e302fd53fb5f8740859d50/stac_fastapi/pgstac/db.py#L48-L62

it would be hard to refactor it in a way to avoid `os.environ` override :-(

lib/titiler-pgstac-api/runtime/src/handler.py
```diff
@@ -8,9 +8,7 @@
 from mangum import Mangum
 from utils import get_secret_dict
 
-pgstac_secret_arn = os.environ["PGSTAC_SECRET_ARN"]
-
-secret = get_secret_dict(pgstac_secret_arn)
+secret = get_secret_dict(secret_arn_env_var="PGSTAC_SECRET_ARN")
```
vincentsarago, line 11: Same here we can define the settings https://github.com/stac-utils/titiler-pgstac/blob/b2aa992554f93acda41473a5eb5e9688ee0a460b/titiler/pgstac/db.py#L13

and pass it to `connect_to_db` function

hrodmn: Yes we could do that but it is impossible to import anything from `titiler.pgstac.main` without the postgres environment variables set because [`assemble_db_connection`](https://github.com/stac-utils/titiler-pgstac/blob/main/titiler/pgstac/settings.py#L77-L90) gets run on every new instance of `PostgresSettings` and it will fail. Another option here would be to set placeholder values in those environment variables in the Lambda configuration then update it in the runtime but that could be a trap for someone who uses a custom runtime.

lib/stac-api/runtime/src/handler.py
hrodmn: maybe we could add a `create_app` factory function to these applications which could be used to create an instance of `app` with some provided settings.

lib/titiler-pgstac-api/runtime/src/handler.py
vincentsarago: I'm going to push a fix and release a new version without the initialization of the settings

lib/database/bootstrapper_runtime/Dockerfile
```diff
@@ -4,8 +4,8 @@ FROM --platform=linux/amd64 public.ecr.aws/lambda/python:${PYTHON_VERSION}
 RUN echo "PYTHON_VERSION: ${PYTHON_VERSION}"
 
 WORKDIR /tmp
-
-RUN pip install httpx psycopg[binary,pool] pypgstac -t /asset
+ARG PGSTAC_VERSION
+RUN pip install httpx psycopg[binary,pool] pypgstac==${PGSTAC_VERSION} -t /asset
```
hrodmn, line 8: I added this because it is possible that you could try to upgrade to a new version of `pypgstac` via migration but if the bootstrapper runtime is running an older version you can't migrate up to a newer version since it is unaware of the new migration path.

lib/database/index.ts
```diff
@@ -14,7 +14,7 @@ import { CustomLambdaFunctionProps } from "../utils";
 import { PgBouncer } from "./PgBouncer";
 
 const instanceSizes: Record<string, number> = require("./instance-memory.json");
-const DEFAULT_PGSTAC_VERSION = "0.8.5";
+const DEFAULT_PGSTAC_VERSION = "0.9.5";
```
hrodmn, line 17: bumping this to get all of the shiny collection search features in the default deployment

lib/stac-api/index.ts
```diff
@@ -23,12 +24,12 @@ export class PgStacApiLambda extends Construct {
...
       runtime: lambda.Runtime.PYTHON_3_11,
-      handler: "src.handler.handler",
+      handler: "handler.handler",
       memorySize: 8192,
       logRetention: aws_logs.RetentionDays.ONE_WEEK,
       timeout: Duration.seconds(30),
-      code: lambda.Code.fromDockerBuild(__dirname, {
-        file: "runtime/Dockerfile",
+      code: lambda.Code.fromDockerBuild(path.join(__dirname, '..'), {
+        file: "stac-api/runtime/Dockerfile",
```
hrodmn, line 32: I changed the docker build path so all of these applications could share a single `utils.py` file.

hrodmn:
I have deployed a less temporary stack with these changes:
stac-fastapi-pgstac: https://mc8hmu6918.execute-api.us-east-1.amazonaws.com/
tipg: https://jnghrxspsi.execute-api.us-east-1.amazonaws.com/
titiler-pgstac: https://dg7bp9uptd.execute-api.us-east-1.amazonaws.com/

Everything seems to be working well with the default applications that are shipped with the tipg/titiler-pgstac/stac-fastapi-pgstac packages. Do these changes warrant a major version release? I think we might be due for v8.0 because of the breaking API changes that we are adding in the default titiler-pgstac and stac-fastapi-pgstac runtimes.

I will see about updating the documentation to help users understand the default behavior and point towards examples of how to customize the applications.

lib/stac-api/runtime/src/handler.py
hrodmn: We refactored the settings in stac-fastapi-pgstac with https://github.com/stac-utils/stac-fastapi-pgstac/pull/209

lib/database/bootstrapper_runtime/Dockerfile
vincentsarago: 👍

PR #131, opened 2025-03-06
fix: use model_dump(mode="json") for serializing STAC objects

### LINKED ISSUE #130: StacCollection.to_dict() cannot be dumped to json with pydantic >= 2
I thought I fixed this in https://github.com/developmentseed/eoapi-cdk/pull/127 but there is more to the story. Items were safe thanks to `jsonable_encoder` but collections were not so lucky. After the upgrade to pydantic>=2 the `AnyUrl` type is not JSON serializable. We weren't catching this for a variety of reasons.

When the ingestor tries to load a collection to the pgstac database, it is hitting a JSON serialization error in the `loader.load_collections` call but it happens silently because of the `try`/`except` logic.
https://github.com/developmentseed/eoapi-cdk/blob/c671ed6acb58a306c04dd3eea088f8890196e116/lib/ingestor-api/runtime/src/collection.py#L17-L26

The exception gets printed by the handler app but never returned to the user via `HTTPException`.

hrodmn:
resolves #130

- [x] Posted the link to a successful manually triggered deployment workflow (successful including the resources destruction) https://github.com/developmentseed/eoapi-cdk/actions/runs/13707782964/job/38337193745

## Merge request description
We did not have coverage for the case where a collection object is not JSON serializable. This fixes a bug with the pydantic>=2 upgrade and adds a test.

### DIFF
--- a/lib/ingestor-api/runtime/src/collection.py
+++ b/lib/ingestor-api/runtime/src/collection.py
@@ -1,5 +1,6 @@
 import os
 
+from fastapi import HTTPException
 from pypgstac.db import PgstacDB
 from pypgstac.load import Methods
 
@@ -18,12 +19,14 @@ def ingest(collection: StacCollection):
         creds = get_db_credentials(os.environ["DB_SECRET_ARN"])
         with PgstacDB(dsn=creds.dsn_string, debug=True) as db:
             loader = Loader(db=db)
-            collection = [
-                collection.to_dict()
-            ]  # pypgstac wants either a string or an Iterable of dicts.
-            loader.load_collections(file=collection, insert_mode=Methods.upsert)
+            loader.load_collections(
+                file=[collection.model_dump(mode="json")], insert_mode=Methods.upsert
+            )
     except Exception as e:
-        print(f"Encountered failure loading collection into pgSTAC: {e}")
+        raise HTTPException(
+            status_code=400,
+            detail=f"Encountered failure loading collection into pgSTAC: {e}",
+        ) from e
 
 
 def delete(collection_id: str):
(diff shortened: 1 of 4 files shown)


lib/ingestor-api/runtime/src/collection.py
```diff
@@ -18,12 +19,14 @@ def ingest(collection: StacCollection):
...
+            )
     except Exception as e:
-        print(f"Encountered failure loading collection into pgSTAC: {e}")
+        raise HTTPException(
+            status_code=400,
+            detail=f"Encountered failure loading collection into pgSTAC: {e}",
+        ) from e
 
 
 def delete(collection_id: str):
```
jjfrench, line 32: Not part of this PR but it looks like the delete func could use some safety catches

lib/ingestor-api/runtime/src/collection.py
```diff
@@ -18,12 +19,14 @@ def ingest(collection: StacCollection):
...
-            collection = [
-                collection.to_dict()
-            ]  # pypgstac wants either a string or an Iterable of dicts.
-            loader.load_collections(file=collection, insert_mode=Methods.upsert)
+            loader.load_collections(
+                file=[collection.model_dump(mode="json")], insert_mode=Methods.upsert
+            )
     except Exception as e:
-        print(f"Encountered failure loading collection into pgSTAC: {e}")
+        raise HTTPException(
```
jjfrench, line 26: Are there additional exceptions we could raise to help debug?  DB_SECRET_ARN could not be set or get_db_credentials could error out

hrodmn: This should catch any error in the whole block and pass a message back to the user. I think it should be a 500 error, though.

lib/ingestor-api/runtime/src/collection.py
hrodmn: we could add some there but at least it would return a 500 if something went wrong there!

jjfrench (approved):

PR #133, opened 2025-03-07
fix: reject items without a collection ID

### LINKED ISSUE #132: check for item collection ID for POST
Right now it is possible to post an item without a collection ID to the ingestor API but it will fail when we try to load it into pgstac because it is missing the collection field.

It would be better to fail on the POST to the API!

hrodmn:
## Merge request description
Right now a user could POST a STAC item without the collection ID field set and the server would hit a 500 error when it tries to find `{stac_endpoint}/collections/None`. This change will instantly reject a STAC item with a missing collection ID!

resolves #132

### DIFF
--- a/lib/ingestor-api/runtime/tests/test_registration.py
+++ b/lib/ingestor-api/runtime/tests/test_registration.py
@@ -93,9 +93,24 @@ class TestCreate:
             collection_id=self.example_ingestion.item.collection
         )
         assert response.status_code == 422, "should get validation error"
-        assert (
-            len(self.db.fetch_many(status="queued")["items"]) == 0
-        ), "data should not be stored in DB"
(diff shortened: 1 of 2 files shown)


lib/ingestor-api/runtime/tests/test_registration.py
```diff
@@ -97,6 +97,21 @@ def test_validates_missing_collection(
             len(self.db.fetch_many(status="queued")["items"]) == 0
         ), "data should not be stored in DB"
 
+    def test_validates_no_collection_id(
+        self, client_authenticated, collection_exists, asset_exists
```
jjfrench, line 101: looks like only `self` is being used

hrodmn: good catch - I just need to include `asset_exists` fixture because apparently that validation check happens before collection ID.

Issue #117, opened 2025-01-22
Add some ops to the integration tests


hrodmn:
After deploying the infrastructure, make some basic requests to make sure the APIs are functional

PR #137, opened 2025-03-11
fix: refactor to keep database pgstac and ingestor pypgstac in sync

hrodmn:
- [x] Posted the link to a successful manually triggered deployment workflow (successful including the resources destruction) https://github.com/developmentseed/eoapi-cdk/actions/runs/13787766250

## Merge request description
It is important for the ingestor API and the database to have the same version of pgstac/pypgstac. This moves the pypgstac version out of a hard-coded value in requirements.txt and into a configurable argument that is easier to keep in sync with the database.

I also added an ingestor deployment to the test deployment CDK app, which should pave the way for #117

### DIFF
--- a/lib/ingestor-api/runtime/Dockerfile
+++ b/lib/ingestor-api/runtime/Dockerfile
@@ -1,4 +1,5 @@
 ARG PYTHON_VERSION
+
 FROM --platform=linux/amd64 public.ecr.aws/lambda/python:${PYTHON_VERSION}
 
 WORKDIR /tmp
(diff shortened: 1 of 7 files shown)


lib/ingestor-api/runtime/Dockerfile
```diff
@@ -8,7 +9,10 @@ RUN yum install -y git
 RUN python -m pip install pip -U
 
 COPY runtime/requirements.txt requirements.txt
-RUN python -m pip install -r requirements.txt "mangum>=0.14,<0.15" -t /asset --no-binary pydantic
+
+ARG PGSTAC_VERSION=0.9.5
```
jjfrench, line 13:
```suggestion
ARG PGSTAC_VERSION
```

We shouldn't need this, right? Since we're passing it through build args?

lib/utils/index.ts
```diff
@@ -1,5 +1,4 @@
-import {
-    aws_lambda as lambda,
-  } from "aws-cdk-lib";
+import { aws_lambda as lambda } from "aws-cdk-lib";
 
 export type CustomLambdaFunctionProps = lambda.FunctionProps | any;
+export const DEFAULT_PGSTAC_VERSION = "0.9.5";
```
jjfrench, line 4: Would a shared requirements file seem too heavy for this usecase?

hrodmn: I think that would be too much since we want users to be able to configure the pgstac version for the database (and now the ingestor)

jjfrench: Managing versions in a utils file seems a little strange to me, but I could see how having it added to constraints keeps the requirements file clean

hrodmn: yeah, I agree in general but this is a simple way to define a default value for the `pgstacVersion` parameters from multiple constructs in a single place!

jjfrench (approved):
might be worth revisiting if we end up abstracting more individual dependencies, but this seems good 👍🏼

PR #142, opened 2025-04-15
refactor: permit usage of higher CDK versions

### LINKED ISSUE #141: update aws-cdk-lib version
we have this package pinned to aws-cdk-lib==2.130.0, current latest is 2.189.0. Could/should we be more flexible in the version constraints somehow? Or should we just bump it to the latest version and pin it again.

alukach:

> [!NOTE]
> Successful build: https://github.com/developmentseed/eoapi-cdk/actions/runs/14480817992/job/40617191146

## Merge request description

I am trying to use `eoapi-cdk` on a project but find am unable to upgrade CDK beyond `2.130.0`:

```
▶ uv sync
  × No solution found when resolving dependencies:
  ╰─▶ Because only eoapi-cdk<=8.0.2 is available and eoapi-cdk==8.0.2 depends on aws-cdk-lib==2.130.0, we can conclude that eoapi-cdk>=8.0.2 depends on
      aws-cdk-lib==2.130.0.
      And because my-project-aws depends on aws-cdk-lib>=2.189.0, we can conclude that -project-aws and eoapi-cdk>=8.0.2 are incompatible.
      And because -project-aws depends on eoapi-cdk>=8.0.2 and your workspace requires -project-aws, we can conclude that your
      workspace's requirements are unsatisfiable.
```

I would prefer to use the latest version of CDK (`2.189.0`).

This PR refactors the dependency declarations in `package.json` to follow best practices for CDK construct libraries[^1]:

> In `peerDependencies`, use a caret (^) to specify the lowest version of `aws-cdk-lib` that your library works with. This maximizes the compatibility of your library with a range of CDK versions. Specify exact versions for alpha construct library modules, which have APIs that may change. Using `peerDependencies` makes sure that there is only one copy of all CDK libraries in the `node_modules` tree.

> In `devDependencies`, specify the tools and libraries you need for testing, optionally with ^ to indicate that later compatible versions are acceptable. Specify exactly (without ^ or ~) the lowest versions of `aws-cdk-lib` and other CDK packages that you advertise your library be compatible with. This practice makes sure that your tests run against those versions. This way, if you inadvertently use a feature found only in newer versions, your tests can catch it.

[^1]: https://docs.aws.amazon.com/cdk/v2/guide/work-with-cdk-typescript.html#work-with-cdk-typescript-dependencies-libraries

closes #141

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,8 +56,8 @@
     "semantic-release": "^19.0.5"
   },
   "peerDependencies": {
-    "@aws-cdk/aws-apigatewayv2-alpha": "^2.114.1-alpha.0",
-    "@aws-cdk/aws-apigatewayv2-integrations-alpha": "^2.114.1-alpha.0",
+    "@aws-cdk/aws-apigatewayv2-alpha": "2.114.1-alpha.0",
+    "@aws-cdk/aws-apigatewayv2-integrations-alpha": "2.114.1-alpha.0",
     "aws-cdk-lib": "^2.130.0",
     "constructs": "^10.3.0"
   },
--- a/package.json
+++ b/package.json
@@ -39,31 +39,27 @@
     }
   },
   "devDependencies": {
+    "@aws-cdk/aws-apigatewayv2-alpha": "2.114.1-alpha.0",
+    "@aws-cdk/aws-apigatewayv2-integrations-alpha": "2.114.1-alpha.0",
     "@qiwi/semantic-release-gh-pages-plugin": "^5.2.3",
     "@semantic-release/changelog": "^6.0.1",
     "@semantic-release/git": "^10.0.1",
     "@types/node": "^18.7.23",
-    "jsii": "5.7.4",
+    "aws-cdk-lib": "2.130.0",
+    "constructs": "10.3.0",
     "jsii-docgen": "10.6.1",
     "jsii-pacmak": "1.106.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -39,31 +39,27 @@
...
-    "@aws-cdk/aws-apigatewayv2-alpha": "2.114.1-alpha.0",
-    "aws-cdk-lib": "2.130.0",
-    "constructs": "10.3.0"
+    "semantic-release": "^19.0.5"
   },
   "peerDependencies": {
-    "aws-cdk-lib": "2.130.0",
-    "constructs": "10.3.0",
-    "@aws-cdk/aws-apigatewayv2-alpha": "2.114.1-alpha.0"
+    "@aws-cdk/aws-apigatewayv2-alpha": "^2.114.1-alpha.0",
```
hrodmn, line 59: we want to pin to a specific version of the `alpha` dependencies, right?

hrodmn (commented):
Thanks for making this change! I ran into the same issue and filed #141 to deal with it but you beat me to it. The only possible change we might need is to pin to a specific version of the `*-alpha` packages (based on the linked docs).

package.json
alukach: Pinned in dd7281dcaea058fcee392ce2b2fdef98c2684290

hrodmn (approved):

PR #146, opened 2025-04-18
chore: update deployment workflow to support pull requests and configure AWS credentials

alukach:
## Merge request description

- Added support for pull request triggers in the deployment workflow for specific paths.
- Configured AWS credentials using a role to assume, enhancing security.
- Updated the AWS region to 'us-west-2' and removed hardcoded access keys.

### DIFF
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -2,18 +2,31 @@ name: Deployment
 
 on:
   workflow_dispatch:
+  pull_request:
+    paths:
+      - "lib/**"
+      - "integration_tests/**"
+      - "package.json"
+      - "package-lock.json"
 
 jobs:
   build_package_and_deploy:
     name: Build, package and deploy
     runs-on: ubuntu-latest
     timeout-minutes: 90
+    permissions:
+      id-token: write
+      contents: read
     env:
-      AWS_DEFAULT_REGION: ${{ secrets.AWS_DEFAULT_REGION_DEPLOY }}
-      AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID_DEPLOY }}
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy.yaml
```diff
@@ -2,18 +2,32 @@ name: Deployment
 
 on:
   workflow_dispatch:
+  pull_request:
+    paths:
+      - "lib/**"
+      - "integration_tests/**"
```
hrodmn, line 8: why would we run the deployment workflow if we just update the integration test?

hrodmn (commented):
Maybe consider removing `integration_tests/**` from the workflow trigger paths - otherwise looks great!

.github/workflows/deploy.yaml
hrodmn: ah nevermind, I forgot the integration tests are where we actually run the deployment :facepalm:

hrodmn (approved):

.github/workflows/deploy.yaml
```diff
@@ -2,18 +2,32 @@ name: Deployment
...
+    permissions:
+      id-token: write
+      contents: read
     env:
-      AWS_DEFAULT_REGION: ${{ secrets.AWS_DEFAULT_REGION_DEPLOY }}
-      AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID_DEPLOY }}
-      AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY_DEPLOY }}
+      AWS_ROLE_ARN: ${{ vars.AWS_ROLE_ARN }}
       AWS_DEFAULT_ACCOUNT: ${{ secrets.AWS_ACCOUNT_ID }}
+      AWS_DEFAULT_REGION: 'us-west-2'
```
hrodmn, line 23: We might keep this in `us-east-1` or any region but `us-west-2` to avoid bumping into resource quotas.

PR #148, opened 2025-04-30
fix: upgrade pgbouncer ami to ubuntu 24.04

hrodmn:

I also manually deployed a stack and verified that the pgbouncer service is working correctly after this change.

## Merge request description
This helps reduce security vulnerabilities for the pgbouncer EC2 instance by upgrading to Ubuntu 24.04 and running `apt-get update` and `apt-get upgrade` to ensure that the latest packages are installed when the instance boots up.


jjfrench (approved):
👍🏼

PR #150, opened 2025-05-28
feat: add stactools-item-generator and stac-item-loader constructs

hrodmn:
- [x] Posted the link to a successful manually triggered [deployment workflow](https://github.com/developmentseed/eoapi-cdk/actions/runs/15349535042/job/43193959748) (successful including the resources destruction)

## Merge request description

There has been appetite for an event-based STAC item generation/ingestion infrastructure in several recent projects (e.g. NASA MAAP). I put together a POC of a SNS/SQS based system over in https://github.com/developmentseed/stactools-ingest but in the interest of making it available via eoapi-cdk constructs I thought I would move it over here for now.

This PR adds two new constructs:

- `StactoolsItemGenerator`: a SNS/SQS/Lambda combo that can use any stactools package to generate a STAC item
- `StacItemLoader`: a SNS/SQS/Lambda combo that loads batches of STAC items into a pgstac database

The `StacItemLoader` is designed to be useful for any pipeline that could publish STAC metadata to an SNS topic or upload STAC items to an S3 bucket (not just `StacItemGenerator`).

Ref: https://github.com/developmentseed/eoAPI/discussions/211


lib/stac-item-generator/runtime/src/stac_item_generator/handler.py
gadomski, line 1: FWIW this is pretty close to https://github.com/stac-utils/stac-task, but I like this better as it's more procedural. **stac-task** did the whole inheritance thing which I find pretty unmaintainable.

gadomski (commented):
Forgive if I'm missing something, but it looks like the loader is set up to take from SNS _or_ S3 notifications, but the generator just goes to SNS? Was there a pre-baked S3 path for the generator that I missed?

alukach:
"Known unknowns" for documentation:

1. What happens with items in the DLQ is not solved
1. Drift detection between S3 and STAC catalog to detect possible S3 -> SNS -> SQS hand-off failures
1. "Fast failure" for synchronous validation of items or auth scenarios could take place in an API that would validate and write successful items to S3

hrodmn:
A few more notes:
- maybe add some logic to drop bad items from a failed `pypgstac load` and retrying good ones to avoid good items being relegated to the DLQ
- don't try to handle the DLQ yet - let users do that independently for now
- the `StacItemLoader` is the most valuable feature, the `StacItemGenerator` could work for some use-cases but we should pursue something that is easier to make generic tooling for (protocol-based).

hrodmn:
I have made a few changes in the PR today after testing some batch STAC item processing (uploading ~1000 STAC items to an S3 bucket):
- Throttled the `StacItemLoader` Lambda and SQS batcher to a maximum concurrency of 2 in order to avoid many concurrent `pypgstac load` commands being run simultaneously. I never hit this in my testing but I assume it could happen so it would be a safe choice to limit the parallelism a bit here. This behavior can be controlled with the `maxConcurrency` prop in `StacItemLoaderProps` cc @sharkinsspatial 
- `StacItemGenerator` -> `StactoolsItemGenerator` (to leave room for some more generic solution down the road) cc @gadomski 
- Additional details about the `DeadLetterQueue` being a dead-end that users will want to manage somehow! cc @alukach

ceholden:
@hrodmn for the limit on loader concurrency,

>Throttled the StacItemLoader Lambda and SQS batcher to a maximum concurrency of 2 in order to avoid many concurrent pypgstac load commands being run simultaneously. I never hit this in my testing but I assume it could happen so it would be a safe choice to limit the parallelism a bit here.

+1, this seems like a good idea to me. In CSDA we had a similar setup that was running 10x concurrent Lambda functions for ingesting Items from one (1) STAC Collection. This caused a lot of table locks which not only slowed down the ingesters, but also resulted in failures for some of our user-facing PgSTAC consumers that relied on `/search` or Item Collections endpoints. I documented the signals (e.g., RDS PerfInsights showing lock contention), impact, and resolution here, https://github.com/NASA-IMPACT/csda-project/issues/1190

The potential caveat is that the concern about table lock contention wouldn't really apply as long as each Lambda is ingesting Items from different Collections. For very high throughput systems a global `n=2` might be restrictive.

For this case we could expect different Collections to write to different buckets or different bucket prefixes, and then have separate SQS queues + Lambdas that subscribe to the bucket notifications. It's probably easiest to have a separate bucket for very large use cases, but if we wanted just one bucket we can write Items to separate prefixes and subscribing using non-overlapping SNS filters. For example, the forthcoming NISAR mission might have its own ingestion pathway that filters the SNS->SQS subscription to `s3://stac-items/NISAR/`. Every other Collection might share just one queue+ingester and subscribe using an "[anything-but](https://docs.aws.amazon.com/sns/latest/dg/string-value-matching.html#string-anything-but-matching)" filter on the bucket prefix to ignore the "NISAR" prefix. I have some examples of the filter rules in [this part of an ADR](https://github.com/NASA-IMPACT/csdap-stac-api/pull/235/files#diff-1afb340532e45b378e93a1a9b28267d18ee6fd9916f5cd10fa08a00d013016d4R223-R225).

hrodmn:
> The potential caveat is that the concern about table lock contention wouldn't really apply as long as each Lambda is ingesting Items from different Collections. For very high throughput systems a global `n=2` might be restrictive.

I agree that n=2 is not ideal but I wouldn't mind as much if the item processing code were a bit faster. Right now it takes ~2.5 minutes to process a batch of 500 STAC items when the messages are delivered as S3 event notifications :disappointed:. Maybe we could speed that up with some async logic in the loop that processes the messages.

> For this case we could expect different Collections to write to different buckets or different bucket prefixes, and then have separate SQS queues + Lambdas that subscribe to the bucket notifications. It's probably easiest to have a separate bucket for very large use cases, but if we wanted just one bucket we can write Items to separate prefixes and subscribing using non-overlapping SNS filters. For example, the forthcoming NISAR mission might have its own ingestion pathway that filters the SNS->SQS subscription to `s3://stac-items/NISAR/`. Every other Collection might share just one queue+ingester and subscribe using an "[anything-but](https://docs.aws.amazon.com/sns/latest/dg/string-value-matching.html#string-anything-but-matching)" filter on the bucket prefix to ignore the "NISAR" prefix. I have some examples of the filter rules in [this part of an ADR](https://github.com/NASA-IMPACT/csdap-stac-api/pull/235/files#diff-1afb340532e45b378e93a1a9b28267d18ee6fd9916f5cd10fa08a00d013016d4R223-R225).

@ceholden these are all really helpful suggestions. For high throughput systems it would make sense to deploy multiple `StacItemLoaders` that each handle items from a specific set of collections (or just one collection). I think that's the nice thing about having this construct in eoapi-cdk! It will be possible to deploy a fleet of these things each with their own special subscriptions.

PR #167, opened 2025-07-17
feat: Expose API runtime constructs

### LINKED ISSUE #166: How to handle Private APIs?
Many of eoAPI's constructs assume that we are looking to create a publicly available API to serve data via Lambda functions. 

<details>

<summary>Examples...</summary>

https://github.com/developmentseed/eoapi-cdk/blob/4697ce89c282a4affbe0d1663b9f44850c3427b3/lib/stac-api/index.ts#L104-L137
https://github.com/developmentseed/eoapi-cdk/blob/4697ce89c282a4affbe0d1663b9f44850c3427b3/lib/tipg-api/index.ts#L54-L82
https://github.com/developmentseed/eoapi-cdk/blob/4697ce89c282a4affbe0d1663b9f44850c3427b3/lib/ingestor-api/index.ts#L231-L253
https://github.com/developmentseed/eoapi-cdk/blob/4697ce89c282a4affbe0d1663b9f44850c3427b3/lib/titiler-pgstac-api/index.ts#L77-L105

</details>

However, what about scenarios when this is _not_ true? For example, it seems entirely reasonable that the ingestion-api would only be available to services within the VPC.  Alternatively, for something like the STAC Auth Proxy, we will explicitly NOT want the STAC API to be publicly available, instead we will using it as a private service only accessible to the STAC Auth Proxy.

**How can we handle private APIs within eoAPI-CDK?**

### Sub-Constructs

My initial idea is to break up the existing constructs into separate sub-constructs representing the Lambdas and the API Gateways. We could still retain the Lambdas we currently have, but also more easily allow the utilization of targeted sub-constructs when we need to alter the API.

#### Example: `PgStacApiLambda` Before

A singular construct with both the runtime and the API:

https://github.com/developmentseed/eoapi-cdk/blob/4697ce89c282a4affbe0d1663b9f44850c3427b3/lib/stac-api/index.ts#L38-L138

#### Example: `PgStacApiLambda` After

A singular construct that calls out to sub-constructs for the runtime and API:

```ts
// Composite Construct
export class PgStacApiLambda extends Construct {
  readonly url: string;
  public stacApiLambdaFunction: lambda.Function;

  constructor(scope: Construct, id: string, props: PgStacApiLambdaProps) {
    super(scope, id);

    const runtime = new PgStacApiLambdaRuntime(scope, 'runtime', {
      vpc: props.vpc,
      subnetSelection: props.subnetSelection,
      db: props.db,
      dbSecret: props.dbSecret,
      enabledExtensions: props.enabledExtensions,
      apiEnv: props.apiEnv,
      lambdaFunctionOptions: props.lambdaFunctionOptions,
    });
    this.stacApiLambdaFunction = runtime.stacApiLambdaFunction;

    const api = new PgStacApiLambdaApiGateway(scope, 'api', {
      lambdaFunction: runtime.lambdaFunction,
      stacApiDomainName: props.stacApiDomainName,
    });

    this.url = api.url;
  }
}
export type PgStacApiLambdaProps = PgStacApiLambdaRuntimeProps && PgStacApiLambdaApiGatewayProps;

// Lambda Runtime
export class PgStacApiLambdaRuntime extends Construct {
  constructor(scope: Construct, id: string, props: PgStacApiLambdaRuntimeProps) {
      // ...
  }
}

export interface PgStacApiLambdaRuntimeProps {
  // ...
};

// Api Gateway
export class PgStacApiLambdaApiGateway extends Construct {
  constructor(scope: Construct, id: string, props: PgStacApiLambdaApiGatewayProps) {
      // ...
  }
}

export interface PgStacApiLambdaApiGatewayProps {
  // ...
};
```

alukach:
In this PR, we create separate runtime constructs (ie the Lambda functions) from the API Gateway construct.  We still maintain the higher level constructs that have both the runtime and API Gateway, but it's now possible to import and reuse just the runtime construct which bringing a custom API Gateway (e.g. for situations when a user would want a _private_ API Gateway, which must be a REST API rather than an HTTP API).

> [!WARNING]
> This update will change the internal node IDs of the constructs. This will cause the Lambdas and API Gateways to be redeployed when upgrading eoapi-cdk. Being that both services are stateless, this should not cause issues unless a developer has made manual modifications or manual references to the deployed lambdas/api gateways.

closes #166

### DIFF
--- a/lib/stac-api/index.ts
+++ b/lib/stac-api/index.ts
@@ -8,11 +8,7 @@ import {
   aws_logs,
 } from "aws-cdk-lib";
 import { Construct } from "constructs";
-import {
-  CustomLambdaFunctionProps,
-  LambdaApiGateway,
-  LambdaApiGatewayProps,
-} from "../utils";
+import { CustomLambdaFunctionProps, LambdaApiGateway } from "../utils";
 import * as path from "path";
 
 export const EXTENSIONS = {
@@ -168,7 +164,7 @@ export class PgStacApiLambda extends Construct {
 
     const api = new LambdaApiGateway(this, "stac-api", {
       lambdaFunction: runtime.stacApiLambdaFunction,
-      domainName: props.domainName,
+      domainName: props.stacApiDomainName,
       outputName: "url",
     });
 
@@ -176,6 +172,11 @@ export class PgStacApiLambda extends Construct {
   }
 }
 
-export interface PgStacApiLambdaProps
-  extends PgStacApiLambdaRuntimeProps,
-    Omit<LambdaApiGatewayProps, "lambdaFunction"> {}
+export interface PgStacApiLambdaProps extends PgStacApiLambdaRuntimeProps {
+  /**
+   * Custom Domain Name Options for Titiler Pgstac API,
+   *
+   * @default - undefined.
+   */
+  readonly stacApiDomainName?: apigatewayv2.IDomainName;
+}
--- a/lib/stac-api/index.ts
+++ b/lib/stac-api/index.ts
@@ -1,17 +1,18 @@
 import {
-  Stack,
   aws_apigatewayv2 as apigatewayv2,
-  aws_apigatewayv2_integrations as apigatewayv2_integrations,
   aws_ec2 as ec2,
   aws_rds as rds,
   aws_lambda as lambda,
   aws_secretsmanager as secretsmanager,
-  CfnOutput,
   Duration,
   aws_logs,
 } from "aws-cdk-lib";
 import { Construct } from "constructs";
-import { CustomLambdaFunctionProps } from "../utils";
+import {
+  CustomLambdaFunctionProps,
+  LambdaApiGateway,
+  LambdaApiGatewayProps,
+} from "../utils";
 import * as path from "path";
 
 export const EXTENSIONS = {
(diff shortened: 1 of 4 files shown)


lib/stac-api/index.ts
```diff
@@ -100,44 +100,10 @@ export class PgStacApiLambda extends Construct {
...
-        ),
-      }
-    );
-
-    this.url = stacApi.url!;
-
-    new CfnOutput(this, "stac-api-output", {
-      exportName: `${Stack.of(this).stackName}-url`,
-      value: this.url,
-    });
```
hrodmn, line 136: I see that you have axed the `CfnOutputs` from these constructs - that's fine with me but just want to make sure it is intentional. I think leaving that up to the user to include in their CDK app is for the best anyways!

hrodmn (approved):
Thanks for doing this reafactor @alukach! I think it is a really useful addition. When I was [experimenting with Lambda SnapStart](https://github.com/MAAP-Project/maap-eoapi/pull/56/files#diff-a38af3cfbbb9a1aec18d30037cab4e9ad0ff7e466c0595e721875a065f507145R210-R220) it occurred to me that it would be nice to have more control over the API settings - this PR would make that a lot easier!

This probably warrants a new major version because of the potentially hazardous renaming of Lambdas and API Gateway APIs.

The only other change I would consider would be to add an example of how to use the new constructs to set up a private API in the docs.

lib/stac-api/index.ts
alukach: I moved them into the API Gateway construct:

https://github.com/developmentseed/eoapi-cdk/blob/0c9d5a0d2bbbe0ccb5ff8488b220983603d39b69/lib/lambda-api-gateway/index.ts#L68-L71

I'm now questioning if that was the right call...

PR #171, opened 2025-07-18
feat: add stac-auth-proxy construct

### LINKED ISSUE #170: Support STAC Auth Proxy
The STAC Auth Proxy[^proxy] should be made available via eoAPI CDK.  In efforts to conform with the remainder of the eoAPI, it would likely make sense to first support a Lambda runtime within the proxy (https://github.com/developmentseed/stac-auth-proxy/issues/60).

[^proxy]: https://github.com/developmentseed/stac-auth-proxy

alukach:
currently available at https://pypi.org/project/eoapi-cdk/9.2.0rc3/

closes #170

### DIFF
--- a/lib/stac-auth-proxy/index.ts
+++ b/lib/stac-auth-proxy/index.ts
@@ -1,0 +1,75 @@
+import * as cdk from "aws-cdk-lib";
+import * as certificatemanager from "aws-cdk-lib/aws-certificatemanager";
+import * as ec2 from "aws-cdk-lib/aws-ec2";
+import * as ecs from "aws-cdk-lib/aws-ecs";
+import * as ecs_patterns from "aws-cdk-lib/aws-ecs-patterns";
+import * as elbv2 from "aws-cdk-lib/aws-elasticloadbalancingv2";
+import { Construct } from "constructs";
+
+export interface StacAuthProxyProps {
+  vpc: ec2.IVpc;
+  upstreamUrl: string;
+  oidcDiscoveryUrl: string;
+  stacApiClientId: string;
+  subnetSelection: ec2.SubnetSelection;
+  certificateArn?: string;
+}
+
+export class StacAuthProxy extends Construct {
+  public readonly service: ecs_patterns.ApplicationLoadBalancedFargateService;
+
+  constructor(scope: Construct, id: string, props: StacAuthProxyProps) {
+    super(scope, id);
+
+    const taskOptions: ecs_patterns.ApplicationLoadBalancedTaskImageOptions = {
+      image: ecs.ContainerImage.fromRegistry(
+        "ghcr.io/developmentseed/stac-auth-proxy:0.5.0"
+      ),
+      containerPort: 8000,
+      environment: {
+        // stac-auth-proxy config
+        UPSTREAM_URL: props.upstreamUrl,
+        OIDC_DISCOVERY_URL: props.oidcDiscoveryUrl,
+        DEFAULT_PUBLIC: "false",
+        OPENAPI_SPEC_ENDPOINT: "/api",
+        // swagger-ui config
+        SWAGGER_UI_ENDPOINT: "/api.html",
+        SWAGGER_UI_INIT_OAUTH: JSON.stringify({
+          clientId: props.stacApiClientId,
+          usePkceWithAuthorizationCodeGrant: true,
+        }),
+      },
+    };
+
+    const albCertificate = props.certificateArn
+      ? certificatemanager.Certificate.fromCertificateArn(
+          this,
+          "cert",
+          props.certificateArn
+        )
+      : undefined;
+
+    this.service = new ecs_patterns.ApplicationLoadBalancedFargateService(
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "eoapi-cdk",
-  "version": "9.1.1",
+  "version": "9.2.0-rc.3",
```
alukach, line 3:
```suggestion
  "version": "9.1.1",
```

Needed to set this for publishing release candidates...

lib/stac-auth-proxy/index.ts
```diff
@@ -0,0 +1,138 @@
...
+
+  constructor(
+    scope: Construct,
+    id: string,
+    props: StacAuthProxyLambdaRuntimeProps
+  ) {
+    super(scope, id);
+
+    this.lambdaFunction = new lambda.Function(this, "lambda", {
+      runtime: lambda.Runtime.PYTHON_3_13,
```
gadomski, line 21: High-level question for **eoAPI-cdk** ... could/should we use a common Python runtime for our lambdas (aka a package-level constant)?

Feels like that could be a _slight_ maintenance burden reduction.

lib/stac-auth-proxy/runtime/Dockerfile
```diff
@@ -0,0 +1,13 @@
+ARG PYTHON_VERSION
+FROM --platform=linux/amd64 public.ecr.aws/lambda/python:${PYTHON_VERSION}
+
+WORKDIR /tmp
+RUN python -m pip install pip -U
+
+COPY stac-auth-proxy/runtime/requirements.txt requirements.txt
+RUN python -m pip install -r requirements.txt "mangum>=0.14,<0.15" -t /asset
```
gadomski, line 8: Feels a little strange to specify dependencies here _and_ in `requirements.txt`?

lib/stac-auth-proxy/runtime/Dockerfile
```diff
@@ -0,0 +1,13 @@
+ARG PYTHON_VERSION
+FROM --platform=linux/amd64 public.ecr.aws/lambda/python:${PYTHON_VERSION}
+
+WORKDIR /tmp
+RUN python -m pip install pip -U
```
gadomski, line 5: If we used **uv** we'd get dependency locking for (almost) free ... is this something we want (I kind of think yes, tracking downstream issues might be easier if we lock our dependency tree)?

lib/stac-auth-proxy/runtime/Dockerfile
```diff
@@ -0,0 +1,13 @@
...
+
+COPY stac-auth-proxy/runtime/requirements.txt requirements.txt
+RUN python -m pip install -r requirements.txt "mangum>=0.14,<0.15" -t /asset
+
+RUN mkdir -p /asset/src
+COPY stac-auth-proxy/runtime/src/*.py /asset/
+
+CMD ["echo", "hello world"]
```
gadomski, line 13: This might be :old_man_yells_at_cloud:, but these (throwaway `CMD`) always feel icky to me in Dockerfiles. Do we need it? Can we populate it with something useful>

lib/stac-auth-proxy/index.ts
alukach: I think that's reasonable. It's hard for me to imagine a situation where upgrading Python would break a service.  I'll go even a step further and suggest that it would be reasonable to have a common _default_ but allow end-users to specify the environment.

gadomski: My original comment was brought to you by maintenance of https://earth-search.aws.element84.com/v1, which seemed to run into Lambda env deprecations all the time which always managed to be harder to unpick than we wanted. I figure having a single source of version helps us catch bitrot earlier and (hopefully) make the upgrades less painful.

lib/stac-auth-proxy/runtime/Dockerfile
alukach: Looks like the world continues to spin if we remove it (ff80958)

lib/stac-auth-proxy/runtime/Dockerfile
alukach: I like this idea, but I think it would be worthy of a separate PR.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "eoapi-cdk",
-  "version": "9.1.1",
+  "version": "9.2.0-rc.8",
```
gadomski, line 3: Seems strange to jump to rc.8? Don't know how packaging/releasing works for this repo, though.

gadomski (approved):

package.json
alukach: Yeah, I've been setting rc and then issuing pre-releases of this branch to test this with a project.  Will revert back to `9.1.1`, CICD should set the version properly.

alukach:
```suggestion
  "version": "9.1.1",
```

PR #214, opened 2025-12-01
chore: update node lambda runtime to latest

gadomski:
- [x] Posted the link to a successful manually triggered deployment workflow (successful including the resources destruction): https://github.com/developmentseed/eoapi-cdk/actions/runs/19827322427

## Merge request description

A client has reported that AWS was yelling about deprecations on node 20 runtimes, so I figure sticking to latest would be good?

### DIFF
--- a/lib/ingestor-api/runtime/dev_requirements.txt
+++ b/lib/ingestor-api/runtime/dev_requirements.txt
@@ -1,3 +1,4 @@
 httpx
 moto[dynamodb, ssm]>=4.0.9,<5.0
 pytest
+fastapi>= 0.122.0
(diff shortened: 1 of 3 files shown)


hrodmn:
> @hrodmn re-requesting approval, tests failed b/c of the 403 to 401 change in https://github.com/fastapi/fastapi/releases/tag/0.122.0. I've only raised the fastapi version in the dev dependencies of the ingestor, since presumably the older versions still work, they just break our tests.

Thanks @gadomski - I followed a follow-up to use `uv` to manage the ingestor-api so we can do a better job avoiding future dependency grenades https://github.com/developmentseed/eoapi-cdk/issues/215

PR #221, opened 2025-12-18
fix: add secretBootstrapper as a dependency for lambda functions

hrodmn:
- [x] Posted the link to a successful manually triggered deployment workflow (successful including the resources destruction)  https://github.com/developmentseed/eoapi-cdk/actions/runs/20351912526

## Merge request description
Avoids race condition between SnapStart cache and pgbouncer secret.

Sometimes, the pgbouncer secret would not be updated at the time that the Lambda SnapStart versions were cached which led to the function being unable to connect to the database.

### DIFF
--- a/lib/utils/index.ts
+++ b/lib/utils/index.ts
@@ -72,7 +72,7 @@ export function extractDatabaseDependencies(db: any): CustomResource[] {
 }
 
 /**
- * Creates a Lambda version with optional dependencies for SnapStart.
+ * Creates a Lambda version with dependencies for SnapStart.
  *
  * When SnapStart is enabled, the version creation triggers snapshot creation.
  * Dependencies ensure the snapshot isn't created until prerequisites are met,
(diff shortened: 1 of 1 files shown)


lib/utils/index.ts
```diff
@@ -15,17 +15,88 @@ export const DEFAULT_PGSTAC_VERSION = "0.9.5";
...
+ * be created until the secret contains the correct connection information.
+ *
+ * @param db - Database construct (may be PgStacDatabase, IDatabaseInstance, etc.)
+ * @returns Array of CustomResources to depend on (empty if none available)
+ */
+export function extractDatabaseDependencies(db: any): CustomResource[] {
+  if (hasSecretBootstrapper(db)) {
+    return [db.secretBootstrapper];
+  }
+  return [];
```
jjfrench: Just checking, so if SnapStart isn't set then we don't necessarily need db.secretBootstrapper?

PR #262, opened 2026-05-27
feat: upgrade API runtime dependencies

hrodmn:
- [x] Posted the link to a successful manually triggered deployment workflow (successful including the resources destruction) https://github.com/developmentseed/eoapi-cdk/actions/runs/26583405654/job/78322530095

## Merge request description
Run `uv sync --upgrade` in all of the runtime directories to upgrade all dependencies.

Upgrading starlette and fastapi required some changes to the handler functions. Since I was working on that I unpinned us from mangum==0.19.0 so now we can stay up to date with new versions.

I also upgraded stac-auth-proxy to v1.1.1 to incorporate some security updates (supersedes #263).


.github/workflows/deploy.yaml
gadomski, line 1: We're pot-committed on the pattern, but I don't love having big bash scripts in Github workflows. I don't think it's worth fixing on this PR, but _to me_ it'd be more maintainable to have this as standalone scripts that we can call from the workflow.

lib/ingestor-api/runtime/src/handler.py
```diff
@@ -1,9 +1,32 @@
...
+        return asyncio.get_running_loop()
+    except RuntimeError:
+        pass
+
+    try:
+        return asyncio.get_event_loop()
+    except RuntimeError:
+        loop = asyncio.new_event_loop()
+        asyncio.set_event_loop(loop)
+        return loop
```
gadomski, line 23: Can you speak a bit more to why we need this? Do lambdas not always have a running loop? Is this a snap start thing?

gadomski (commented):
👍🏼 to allowing **mangum** version to float. It does feel like we're duplicating a lot of kinda complex code — is there any path to DRY-ing it up?

.github/workflows/deploy.yaml
hrodmn: Good call, I filed a follow-up ticket: https://github.com/developmentseed/eoapi-cdk/issues/264

lib/ingestor-api/runtime/src/handler.py
hrodmn: I copied the pattern from the changes that I made in the stac-fastapi-pgstac handler, but since this handler does not use SnapStart (yet) maybe it's not necessary.

The changes in the SnapStart-enabled handlers were necessary because we moved on from the deprecated `@app.on_event("startup")` pattern which handled a bunch of event loop logic under the hood (I think).

gadomski (approved):

hrodmn:
> 👍🏼 to allowing mangum version to float. It does feel like we're duplicating a lot of kinda complex code — is there any path to DRY-ing it up?

We have a mechanism for doing that already in `utils.py` - I'll check to see if any of the common functions could just be defined there and imported in each handler.

PR #281, opened 2026-08-04
fix: add stac-browser v5.x support

jjfrench:

## Merge request description
- Modifies the stac-browser buildapp to support v5.0.0 of the radiantearth stac-browser by checking the tag version and passing props into their respective builds v3/v4,v5

## How you can test it
By commenting out the bucket section of the construct, you can easily target the build portion of the stac-browser construct
```[TypeScript]
// test-build.ts
import * as cdk from 'aws-cdk-lib';

// lib/stac-browser/index.ts
import { StacBrowser } from '../lib/stac-browser/index.ts';


function main() {

    // Throwaway app/stack purely to satisfy the Construct tree — never synthesized
    // to a template file or deployed.
    const app = new cdk.App();
    const stack = new cdk.Stack(app, 'TestStack');

    // Instantiating this runs buildApp() as a side effect of the constructor.
    new StacBrowser(stack, 'TestStacBrowser', {
        stacCatalogUrl: "https://stac.maap-project.org",
        githubRepoTag: "v3.2.0", // v5.0.0
        websiteIndexDocument: 'index.html',
    });

    console.log(`   Try: npx serve ./stac-browser/dist`);
}

main();
```
you might need a tsconfig.json, e.g.
```[json]
{
  "compilerOptions": {
    "target": "ES2022",
    "module": "NodeNext",
    "moduleResolution": "NodeNext",
    "allowImportingTsExtensions": true,
    "strict": false,
    "skipLibCheck": true
  },
  "include": ["./**/*.ts"]
}
```

```npx ts-node test-build.ts ```

```npx serve ./stac-browser/dist```

### DIFF
--- a/lib/stac-browser/index.ts
+++ b/lib/stac-browser/index.ts
@@ -158,7 +158,6 @@ export interface StacBrowserProps {
 
     /**
      * STAC catalog URL. Overrides the catalog URL in the stac-browser configuration.
-     * Backwards compatibility for version prior to v5.0.0.
      */
     readonly stacCatalogUrl: string;
 
@@ -174,8 +173,6 @@ export interface StacBrowserProps {
 
     /**
      * Tag of the radiant earth stac-browser repo to use to build the app.
-     * Supports v3.x, v4.x, and v5.x tags — the build invocation is automatically
-     * adjusted to match the config mechanism supported by that version.
      */
     readonly githubRepoTag: string;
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: lib/stac-browser/index.ts [hunk 1]
```diff
@@ -102,11 +113,32 @@ export class StacBrowser extends Construct {
             fs.copyFileSync(props.configFilePath, `${cloneDirectory}/config.js`);
         }
 
-        // Build the app with catalogUrl
-        console.log(`Building app with catalogUrl=${props.stacCatalogUrl} into ${cloneDirectory}`)
-        execSync(`npm run build -- --catalogUrl=${props.stacCatalogUrl}`, { cwd: cloneDirectory });
+        // Build the app with catalogUrl.
+        // See: https://github.com/radiantearth/stac-browser/discussions/751
+        const majorVersion = StacBrowser.parseMajorVersion(props.githubRepoTag);
```
hrodmn, line 118: do we need `StacBrowser`. here or could we just reference `parseMajorVersion` since it is defined in this module?

hrodmn (approved):
LGTM! Maybe we want to use `fix: add stac-browser v5.x support` if we just want it to be a patch instead of a new minor version. No strong preference from me, just throwing it out there.

lib/stac-browser/index.ts
jjfrench: It's currently a static method because it's not reliant on any instance state but I could remove that and use "this" if that's preferable

PR #288, opened 2026-08-06
fix: increase memory of stac-browser BucketDeployment

jjfrench:

## Merge request description
- Increase the memory limits of the stac-browser BucketDeployment. Initial deployment shows a timeout with near max memory usage
  >Duration: 900000.00 ms Billed Duration: 900000 ms Memory Size: 128 MB Max Memory Used: 126 MB Status: timeout
- Added stac-browser to integration build test
- Added auto_delete_objects for stac-browser s3 bucket

### DIFF
--- a/integration_tests/cdk/app.py
+++ b/integration_tests/cdk/app.py
@@ -13,6 +13,7 @@ from constructs import Construct
 from eoapi_cdk import (
     PgStacApiLambda,
     PgStacDatabase,
+    StacBrowser,
     StacIngestor,
     StacLoader,
     StactoolsItemGenerator,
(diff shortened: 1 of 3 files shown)


hrodmn (commented):
Let's add a STAC Browser construct to the [integration test app](https://github.com/developmentseed/eoapi-cdk/blob/85b0e3314de9dc4a924fa4fb854e6b2046e5bb18/integration_tests/cdk/app.py) so we test the deployment on each PR!

jjfrench:
https://github.com/developmentseed/eoapi-cdk/actions/runs/31118723461/job/92717439125

Will have to take a closer look into the initial deployment failures in the health check, but it seems to have passed this time

jjfrench:
Not entirely sure if the memory issue is reproducible exactly, when we're tearing down the stack each time, but on fresh deployments, the stac-browser deploys successfully in little to no time. 

I've also added the option for auto_delete_objects on the stac-browser bucket - great for our ephemeral stack and user test environments where they may want to delete and redeploy stacks without manually going to empty and delete the bucket.

hrodmn (approved):

